PR #59, opened 2023-04-06
Implement gateway proxy support

### LINKED ISSUE #56: Implement gateway proxy
Booting the bot takes about 5 minutes, waiting for all shards to connect, which is very annoying when applying multiple hotfixes.

[This project](https://github.com/Gelbpunkt/gateway-proxy) fixes the issue by keeping alive the shards for a while, to let the bot resume connection immediately after a reboot.

This could be implemented using docker compose.

PredaaA:
This PR implements the required pieces to get the gateway proxy working on this bot.
It requires a config.json file for the gateway proxy container, a example is included.
Then for the bot to be able to connect using the gateway proxy sessions, the `gateway_url` field must be provided in the config.yml file. In this context it will simply be `gateway_proxy:5421`.

Closes #56.

### DIFF
--- a/ballsdex/__main__.py
+++ b/ballsdex/__main__.py
@@ -88,7 +88,7 @@ def print_welcome():
 
 def patch_gateway(proxy_url: str):
     """This monkeypatches discord.py in order to be able to use a custom gateway URL.
-    
+
     Parameters
     ----------
     proxy_url : str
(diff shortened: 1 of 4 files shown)


PredaaA:
Note that it is also good to know that latencies on the bot will be incorrect due to the heartbeat stuff in dpy. 
The gateway proxy container has a Prometheus endpoint that has all shards latencies if needed to track.

PR #64, opened 2023-04-10
FEAT: Add confirmation check to trading favourite ball.

flaree:
Due to community suggestions, adding a confirmation button to trade when ball is favourited.

First time actually trying to use views, let me know if there's anything wrong.

### DIFF
--- a/ballsdex/core/utils/buttons.py
+++ b/ballsdex/core/utils/buttons.py
@@ -2,6 +2,7 @@ import discord
 
 from discord.ui import View, Button
 
+
 class ConfirmChoiceView(View):
     def __init__(self, interaction: discord.Interaction):
         super().__init__(timeout=90)
@@ -32,7 +33,9 @@ class ConfirmChoiceView(View):
     async def confirm_button(self, interaction: discord.Interaction, button: Button):
         for item in self.children:
             item.disabled = True
-        await interaction.response.edit_message(content=interaction.message.content + '\nConfirmed', view=self)
+        await interaction.response.edit_message(
+            content=interaction.message.content + "\nConfirmed", view=self
+        )
         self.value = True
         self.stop()
 
@@ -43,6 +46,8 @@ class ConfirmChoiceView(View):
     async def cancel_button(self, interaction: discord.Interaction, button: Button):
         for item in self.children:
             item.disabled = True
-        await interaction.response.edit_message(content=interaction.message.content + '\nCancelled', view=self)
+        await interaction.response.edit_message(
+            content=interaction.message.content + "\nCancelled", view=self
+        )
         self.value = False
         self.stop()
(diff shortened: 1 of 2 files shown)


ballsdex/core/utils/buttons.py
```diff
@@ -0,0 +1,39 @@
...
+                "Only the original author can use this.", ephemeral=True
+            )
+            return False
+        if self.value is not None:
+            await interaction.response.send_message(
+                "You've already made a choice.", ephemeral=True
+            )
+            return False
+        return True
+
```
laggron42, line 23: Would suggest adding this to disable the buttons once the view timed out.
```suggestion

    async def on_timeout(self):
        for item in self.children:
            item.disabled = True
        await self.interaction.followup.edit_message("@original", view=self)

```

ballsdex/core/utils/buttons.py
```diff
@@ -0,0 +1,39 @@
...
+    @discord.ui.button(
+        style=discord.ButtonStyle.success, emoji="\N{HEAVY CHECK MARK}\N{VARIATION SELECTOR-16}"
+    )
+    async def confirm_button(self, interaction: discord.Interaction, button: Button):
+        await interaction.response.send_message('Confirming', ephemeral=True)
```
laggron42, line 28: How about editing the original message instead? Since this will be followed up by another message once the command continues, I think seeing 3 messages in a row will be a bit heavy for this.

I also made the buttons disabled since this is editing the message anyway.
```suggestion
        for item in self.children:
            item.disabled = True
        await interaction.response.edit_message(content=interaction.message.content + '\nConfirmed', view=self)
```

ballsdex/core/utils/buttons.py
```diff
@@ -0,0 +1,39 @@
...
+        self.value = True
+        self.stop()
+
+    @discord.ui.button(
+        style=discord.ButtonStyle.danger,
+        emoji="\N{HEAVY MULTIPLICATION X}\N{VARIATION SELECTOR-16}",
+    )
+    async def cancel_button(self, interaction: discord.Interaction, button: Button):
+        await interaction.response.send_message('Cancelling', ephemeral=True)
```
laggron42, line 37: Same as the above suggestion
```suggestion
        for item in self.children:
            item.disabled = True
        await interaction.response.edit_message(content=interaction.message.content + '\nCancelled', view=self)
```

laggron42 (changes requested):
Thanks for the PR!

I like the addition of a new util that could be used by other functions, I will definitely be editing my code to use this new shared function.

Just left a few suggestions, which I haven't tried because it was typed from Github, this should probably need some `black` afterwards.

Speaking of which, PR is not passing the CI, can you run `black` in the repo? You can also setup `pre-commit` to have this done automatically, I explained it [here](https://github.com/laggron42/BallsDex-DiscordBot/blob/master/CONTRIBUTING.md#coding-style)

laggron42 (approved):

PR #65, opened 2023-04-10
feat: add searching to ball instance panel

flaree:
Adds searching on ID, ball ID and player ID to the ball instance panel.

### DIFF
--- a/ballsdex/core/admin/resources.py
+++ b/ballsdex/core/admin/resources.py
@@ -232,17 +232,13 @@ class BallInstanceResource(Model):
         filters.Search(
             name="id",
             label="Ball Instance ID",
-            placeholder="Search for balls",
+            placeholder="Search for ball IDs",
         ),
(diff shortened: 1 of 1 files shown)


ballsdex/core/admin/resources.py
```diff
@@ -228,7 +228,23 @@ class BallInstanceResource(Model):
...
+        ),
+        filters.Search(
+            name="ball_id",
+            label="Ball ID",
+            placeholder="Search for balls",
+        ),
+        filters.Search(
+            name="player",
+            label="Player ID",
+            placeholder="Search for player",
```
laggron42, line 245: You can query relations directly using `__` in some Tortoise fields :P 

I think it would be better if you could search for player Discord ID directly
```suggestion
            name="player__discord_id",
            label="User ID",
            placeholder="Search for Discord user ID",
```

laggron42 (approved):
Thanks for the PR!

Added a few suggested changes I tested locally, but it's mostly suggestions, I leave you free to choose.

PR #66, opened 2023-04-14
BUG: Only show enabled balls in about list

flaree:
Right now there's a chance it'll pick disabled balls and show their emojis, which apparently confuses people


laggron42 (approved):
Can't really go wrong, I'll assume it works

PR #69, opened 2023-04-18
Replace Tortoise ORM with Piccolo ORM

PredaaA:
That's a work in progress, still some things to be done before it's ready for review.

- [ ] Update postgres connection onto the bot and dashboard (dashboard needs to move to another directory).
- [ ] Create a proper app for Piccolo migrations.
- [ ] Update models (might need to move to its own directory too).
- [ ] Update all queries, it will need to use `objects` so we can continue to use class based stuff.


zNeolythic:
is it still gonna happen?

flaree:
No longer migrating to piccolo, closing

PR #70, opened 2023-04-19
feat: Add emoji support for special cards

flaree:
Adds emoji support for special events which will show in trades etc

NOTE: Due to a discord limitation, you cannot show custom emojis in autocomplete/selection models.

Another limitation, the database just stores raw text like `:egg:`, to support showing unicode emojis in those model fields we will need to change handling to convert emoji to its unicode equivalent which is not part of this PR. Instead a generic lightning emoji is used to signify a special card.

### DIFF
--- a/ballsdex/core/models.py
+++ b/ballsdex/core/models.py
@@ -86,7 +86,10 @@ class Special(models.Model):
     democracy_card = fields.CharField(max_length=200)
     dictatorship_card = fields.CharField(max_length=200)
     union_card = fields.CharField(max_length=200)
-    emoji = fields.CharField(max_length=40)
+    emoji = fields.CharField(
+        max_length=20,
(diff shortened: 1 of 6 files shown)


ballsdex/core/models.py
```diff
@@ -85,6 +86,7 @@ class Special(models.Model):
     democracy_card = fields.CharField(max_length=200)
     dictatorship_card = fields.CharField(max_length=200)
     union_card = fields.CharField(max_length=200)
+    emoji = fields.CharField(max_length=40)
```
laggron42, line 89: Added a description and reduced the max length (shouldn't be higher than an ID)
```suggestion
    emoji = fields.CharField(
        max_length=20,
        description="Either a unicode character or a discord emoji ID"
    )
```

laggron42 (changes requested):

PR #71, opened 2023-04-19
BUG: Fix spawn issue related to weight mismatch

flaree:


laggron42 (approved):
Reworked the fix to simply add `[None]` at the end, fixing all problems without conditioning the loops.

PR #73, opened 2023-04-19
FEAT: Add text command for rarity output

flaree:
Adds a text command for a rarity list for continuous or grouped order.

### DIFF
--- a/ballsdex/core/commands.py
+++ b/ballsdex/core/commands.py
@@ -1,6 +1,9 @@
+from collections import defaultdict
 import logging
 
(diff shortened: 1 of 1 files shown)


ltzmax (commented):
did a small test and seems to working. only thing is pre commit failing !

laggron42 (commented):
Can this be made into a slash command under `/admin` instead. I'm trying to avoid text commands in general.

Though I'm not sure how pagination could be managed.

flaree:
I wanted to at first but I have no clue how pagify works in slash so scrapped that idea

PR #79, opened 2023-04-21
Check for ownership validity before transfer

Kowlin:


laggron42 (approved):
Added more explicit end-user error messages, and refresh from DB.

PR #95, opened 2023-05-24
Reset favourite on give

flaree:
Reset favourite on give


ltzmax (approved):
Fully tested and works gud!✅

PR #111, opened 2023-06-04
[Admin] New admin commands

### LINKED ISSUE #83: Add admin commands for objects lookup and operations
Goal is to replace what's being done with eval commands to inspect some stuff. The following things should be done:

- Inspect a ball from its ID
- List and filter balls from player ID or trade player ID
- Change ownership of a ball
- Mass transfer balls of one user to another
- Get a server's information from its ID
- Leave a server
- Get a player's information from its database or Discord ID
- Revert trades (will be done when we have trade history)
- Set bot status

flaree:
New commands that adds the following:

- Ball lookup
- Ball transfer
- Ball delete
- Account reset
- Count of how many

Note: Not tested, not at pc with a ballsdex setup atm

Closes #83

### DIFF
--- a/ballsdex/packages/admin/cog.py
+++ b/ballsdex/packages/admin/cog.py
@@ -33,6 +33,7 @@ class Admin(commands.GroupCog):
     def __init__(self, bot: "BallsDexBot"):
         self.bot = bot
         self.blacklist.parent = self.__cog_app_commands_group__
+        self.balls.parent = self.__cog_app_commands_group__
 
     blacklist = app_commands.Group(name="blacklist", description="Bot blacklist management")
     blacklist_guild = app_commands.Group(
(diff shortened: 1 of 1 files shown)


flaree:
Marking ready to review

ASUSOmen:
When I did implement it on my test bot, even after reloading the tree the commands would not show up. They might be text commands but from looking at the code they look like slash commands. I don't know if I am doing anything wrong but the previous admin commands are not broken and still work. Please tell me if I am doing anything wrong or the code doesn't work yet.

flaree:
> When I did implement it on my test bot, even after reloading the tree the commands would not show up. They might be text commands but from looking at the code they look like slash commands. I don't know if I am doing anything wrong but the previous admin commands are not broken and still work. Please tell me if I am doing anything wrong or the code doesn't work yet.

They're working fine, so it's something on your end

ASUSOmen:
Is b.reloadtree the command that will enable them or do I need to do something else?

flaree:
That's one way, but syncing tree shouldn't really be called manually. Try restarting the bot, if not ping me in the self host server and I'll try help some more

laggron42 (approved):

PR #124, opened 2023-07-01
Fix ball count 

flaree:
Add deferal to ball count to help with large dataset

Correctly show special name in search


ltzmax (approved):
works great!

PR #163, opened 2023-11-11
Remove ability to user lookup on list

flaree:
Removes the User argument on balls list, replaces with Member and adds a transformer error to bot on error to handle stuff that cannot be converted.

### DIFF
--- a/ballsdex/core/bot.py
+++ b/ballsdex/core/bot.py
@@ -419,6 +419,7 @@ class BallsDexBot(commands.AutoShardedBot):
 
             if isinstance(error, app_commands.TransformerError):
                 await send("One of the arguments provided cannot be parsed.")
+                log.debug(f"Failed running converter", exc_info=error.original)
                 return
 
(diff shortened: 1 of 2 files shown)


ballsdex/core/bot.py
```diff
@@ -417,6 +417,10 @@ async def send(content: str):
                 )
                 # still including traceback because it may be a programming error
 
+            if isinstance(error, app_commands.TransformerError):
+                await send("One of the arguments provided cannot be parsed.")
```
laggron42, line 421: I think it's better to log these errors, it's good to acknowledge but errors should still not happen and be handled before.
```suggestion
                await send("One of the arguments provided cannot be parsed.")
                log.error(f"Failed running converter", exc_info=error.original)
```

flaree: I'm not sure if we want to log these as it's just stuff failing the default model converters. I believe our custom transformers handle their own errors

PR #176, opened 2023-12-05
TRADE: Add more tradeable checks

flaree:
Adds tradeable option to BallInstance and Special
Adds `is_tradeable` property to BallInstance to check all 3.


laggron42 (approved):
Tested, fixed a bug, and added the new attributes in admin panel.

PR #184, opened 2023-12-20
Fixed shiny case for balls count

GeorgeSofroniouCy:
Added shiny dictionary for the prompt message, added it in the f string, and also added if balls = 0 for the plural, since it should be "0 countryballs" rather than "0 countryball". First time creating a PR, so hope this works haha, cheers.

### DIFF
--- a/ballsdex/packages/players/cog.py
+++ b/ballsdex/packages/players/cog.py
@@ -539,10 +539,10 @@ class Players(commands.GroupCog, group_name=settings.players_group_cog_name):
         balls = await BallInstance.filter(**filters).count()
         country = f"{countryball.country} " if countryball else ""
         plural = "s" if balls > 1 or balls == 0 else ""
-        shiny = "shiny" if shiny else ""
+        shiny_str = "shiny" if shiny else ""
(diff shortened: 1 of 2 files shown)


ballsdex/packages/players/cog.py
```diff
@@ -538,9 +538,10 @@ async def count(
         await interaction.response.defer(ephemeral=True, thinking=True)
         balls = await BallInstance.filter(**filters).count()
         country = f"{countryball.country} " if countryball else ""
-        plural = "s" if balls > 1 else ""
+        plural = "s" if balls > 1 or balls = 0 else ""
```
flaree, line 541: The = expression on its own is an assignment, you would need == to compare.

An easier way may be to check if the number is equal to one, such as 'plural = "" if balls == 1 else "s"'

GeorgeSofroniouCy: That's right, my bad! That is actually a mistake I often do while coding haha. Let me edix the fork.

laggron42:
Thanks for the contribution!

PR #185, opened 2023-12-20
make it so that you cannot donate or start trade with blacklisted user

MaxxMXSZ:


ballsdex/packages/players/cog.py
```diff
@@ -475,6 +475,10 @@ async def give(
             )
             del self.bot.locked_balls[countryball.pk]
             return
+        if new_player.discord_id in self.bot.blacklist :
```
ltzmax, line 478:
```suggestion
        if new_player.discord_id in self.bot.blacklist:
```

ltzmax (changes requested):

flaree (approved):

laggron42:
Thanks for the contribution!

PR #187, opened 2023-12-24
Added User2 for admin history user

MaxxMXSZ:


flaree:
Thanks for the PR, much appreciated <3

PR #191, opened 2024-01-09
Trade: Handle huge trades

flaree:
Handles huge trades, this is mostly an obscure bug and would be hard to reach otherwise.


ltzmax (approved):
Worked here. i'll wait for @laggron42 incase he wants to look too!

PR #225, opened 2024-03-23
Add atk and hp stat and stat bonus to spawn message

### LINKED ISSUE #140: Include the balls's stats (attack and defence) in the "You caught a ball!" message. (Enhancement)
### Include the balls's stats (attack and defence) in the "You caught a ball!" message.
I think it would be a good idea to show the ball's stats because it would remove the hassle of having to use a command to see the ball's stats.

The message could look like this:

> You caught Andorra! (#1712AEC)
> ATK: 182 (-3%) HP: 411 (-5%)
> 
> [event ball]
> 
> This is a new countryball that has been added to your completion!

MaxxMXSZ:
closes #140

### DIFF
--- a/ballsdex/packages/countryballs/components.py
+++ b/ballsdex/packages/countryballs/components.py
@@ -77,8 +77,8 @@ class CountryballNamePrompt(Modal, title=f"Catch this {settings.collectible_name
                 )
 
             f"{interaction.user.mention} You caught **{self.ball.name}!** "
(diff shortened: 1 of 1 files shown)


flaree:
Honestly not a fan of this, I think it clutters the simplicity of the spawn message

MaxxMXSZ:
Well now its only stat bonus but looks way cleaner!

MaxxMXSZ:
Resolves #140

flaree:
LGTM, thanks for the PR

Issue #222, opened 2024-03-21
Add days argument to trade history


flaree:
Add a days argument to trade history to be able to filter down to trades within the last X days

The Trade model already stores the date so this just encompasses adding a days argument to the trade history AND admin history commands and then using the orm filtering to filter the results effectivelg

MaxxMXSZ:
✔

PR #226, opened 2024-03-24
Added 'day

MaxxMXSZ:
.

### DIFF
--- a/ballsdex/packages/admin/cog.py
+++ b/ballsdex/packages/admin/cog.py
@@ -1342,7 +1342,7 @@ class Admin(commands.GroupCog):
             The sorting method you want to use.
         days: int
             Retrieve ball history for x amount of days, either from the start or the end. 
-            0 = all trade history. Most recent = last x days and Oldest = first x days.
+            0 = all ball history. Most recent = last x days and Oldest = first x days.
         """
 
         try:
--- a/ballsdex/packages/admin/cog.py
+++ b/ballsdex/packages/admin/cog.py
@@ -1246,7 +1246,9 @@ class Admin(commands.GroupCog):
 
         if user2:
             if days < 0:
-                await interaction.followup.send("Invalid number of days. Please provide a non-negative value.", ephemeral=True)
+                await interaction.followup.send(
+                    "Invalid number of days. Please provide a non-negative value.", ephemeral=True
+                )
                 return
 
             if days == 0:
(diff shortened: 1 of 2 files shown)


MaxxMXSZ:
#222

ballsdex/packages/admin/cog.py
```diff
@@ -1235,20 +1237,42 @@ async def history_user(
             The user you want to check the history of.
         sorting: str
             The sorting method you want to use.
+        days: int
+            Retrieve trade history for x amount of days, either from the start or the end.
+            0 = all trade history. Most recent = last x days and Oldest = first x days.
```
laggron42, line 1242: This description is confusing
- The `0` value is already the default value, I think users will understand that not using this argument means getting all history
- The difference you are making between most recent and oldest makes little sense, at least looking at the code there's no difference made in the filtering when one mode or the other is picked. I think you can completely exclude this distinction and instead say something like "retrieve last x days".

ballsdex/packages/trade/cog.py
```diff
@@ -282,20 +284,42 @@ async def history(
...
+            await interaction.followup.send(
+                "Invalid number of days. Please provide a non-negative value.", ephemeral=True
             )
-        else:
+            return
+
+        if days == 0:
             history_queryset = TradeModel.filter(
                 Q(player1__discord_id=user.id) | Q(player2__discord_id=user.id)
             )
```
laggron42, line 305: You removed the condition to use `trade_user` when the days argument is not used. This breaks filtering by user2 in most cases.

laggron42 (changes requested):
Thanks for your work! I tested it and it works well in most cases.

The problem I have is that there's too much duplicated code at this point, the trade history logic is repeated 3 times. I intended to refactor this after your pull request, but this has lead you to mistakes already (forgetting to put back the conditions on user2 in a command).

I'll be following this up on Discord to see what direction to take and help you there.

ballsdex/packages/admin/cog.py
```diff
@@ -1237,39 +1239,45 @@ async def history_user(
             The sorting method you want to use.
         user2: discord.User | None
             The second user you want to check the history of.
+        days: Optional[int]
+            Retrieve trade history from last x day.
```
laggron42, line 1243:
```suggestion
            Retrieve trade history from last x days.
```

ballsdex/packages/admin/cog.py
```diff
@@ -1296,6 +1305,8 @@ async def history_ball(
             The ID of the ball you want to check the history of.
         sorting: str
             The sorting method you want to use.
+        days: Optional[int]
+            Retrieve ball history from last x day.
```
laggron42, line 1309:
```suggestion
            Retrieve ball history from last x days.
```

ballsdex/packages/trade/cog.py
```diff
@@ -284,21 +286,35 @@ async def history(
             The sorting order of the trades
         trade_user: discord.User | None
             The user you want to see your trade history with
+        days: Optional[int]
+            Retrieve trade history from last x day.
```
laggron42, line 290:
```suggestion
            Retrieve trade history from last x days.
```

laggron42 (changes requested):
Niiice thanks for the refactoring!! Just one typo nitpick and we're good to go

laggron42 (approved):

PR #227, opened 2024-03-26
Config tos and policy fix!

MaxxMXSZ:
really small but big issue lol


ltzmax (approved):
Thanks for the pr and the fix!

PR #241, opened 2024-04-23
small fixment to the privacy function

arrayunderscore:
well that should help


ballsdex/packages/balls/cog.py
```diff
@@ -591,10 +591,13 @@ async def inventory_privacy(
         if any(role.id in roles for role in interaction.user.roles):  # type: ignore
             return True
     if privacy_policy == PrivacyPolicy.DENY:
-        await interaction.followup.send(
-            "This user has set their inventory to private.", ephemeral=True
-        )
-        return False
```
flaree, line 597: I'd recommend changing it to just 

```py
if interaction.user.id == player_obj.discord_id:
    return True
```

At the top of the function. If the user invoking it is the same as the provided player then there's no need to run any of the checks needed for privacy policy

flaree (changes requested):

ballsdex/packages/balls/cog.py
arrayunderscore: but how then we will inform that the guy's inventory is private?

flaree: The if check only runs if the user is the invoking command user, in which it just returns that the inventory is not private

arrayunderscore: hmm, alr let me do edits rq

flaree: You didn't read what I said, you need to move that conditional to the start of the method

arrayunderscore: oh, i'm sorry my bad

flaree: Still haven't addressed what I asked, move the if to line 588 and then put the message and return back in the condition for the deny check

arrayunderscore: here made it the first if statement so it doesn't run anything else later on

ballsdex/packages/balls/cog.py
```diff
@@ -586,6 +586,8 @@ async def inventory_privacy(
     user_obj: Union[discord.User, discord.Member],
 ):
     privacy_policy = player.privacy_policy
+    if interaction.user.id == player_obj.discord_id:
```
flaree, line 589: player_obj was just a placeholder in my suggestion, the actual argument is called player. Apologies for that

flaree (changes requested):

flaree (approved):

PR #258, opened 2024-05-19
feat: add user parameter to count

gtm-nayan:
Tiny QoL change for trading commons


flaree:
We are not interested in adding a user arg to count. Please open a suggestion in our discord or create an issue so we can triage first without having to close a PR. 


Thanks for the time

PR #266, opened 2024-05-26
unnecessary space in balls give fix

imtherealF1:
there was an annoying extra space, small fix


flaree (changes requested):
This isn't the actual fix, you need to fix it in the actual model. There's other places where this is an issue. This is just placing some tape over the actual problem

PR #271, opened 2024-06-04
Fix extra space in /ball give

GeorgeSofroniouCy:
I believe this is the root of the extra space, not the actual command in balls cog.


laggron42 (approved):
Thanks!

PR #272, opened 2024-06-04
Fix the annoying space

MaxxMXSZ:
I haven't test it so feel free but from my understanding this should work just fine!


flaree:
duplicate of #271

PR #279, opened 2024-06-10
Add cooldown to /player export

HiboMan:
Prevent users from making the bot spam upload data, with a cooldown of 60 seconds


flaree:
Thanks for the PR, however this command is currently being rewritten as it's not a viable implementation.

PR #281, opened 2024-06-12
fixed /player delete.

Kingofthehill4965:
I saw the bug report in the server that it does not show if the ball is traded if the user who gave that ball deletes their data afterwards. This should fix it since now only the instances the player owns are getting deleted and not all the player data. Please review

### DIFF
--- a/ballsdex/packages/players/cog.py
+++ b/ballsdex/packages/players/cog.py
@@ -101,6 +101,8 @@ class Player(commands.GroupCog):
         """
         Delete your player data.
         """
+        user = interaction.user
+        player = await PlayerModel.get(discord_id=user.id)
         view = ConfirmChoiceView(interaction)
         await interaction.response.send_message(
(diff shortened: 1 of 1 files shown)


laggron42 (changes requested):
Hey, thanks for the contribution!

However, this is not the right way to do things as we *need* the Player object to be deleted for GDPR purposes. What should be done instead is to modify the SQL schema such that deleting a row does not cascade over a trade model.

Sorry, I know it's annoying having to redo things, especially given how painful aerich can be, if you want I can walk you through this, I contacted you on Discord.

Kingofthehill4965:
Hi... Sorry i couldn't respond earlier since I was travelling on vacation. I am free now and can respond as soon as possible

Kowlin:
Pending discussion on #346

flaree:
Hi there, we plan on making a soft delete functionality so I believe this PR will no longer be needed when that comes along.

We appreciate the work put in thus far.

PR #282, opened 2024-06-14
[core] For cooldown show active timer

Buggem:
I imported `time` because `datetime` is much more limiting, and kind of useless. Works with Discord, verified it.

### DIFF
--- a/ballsdex/core/bot.py
+++ b/ballsdex/core/bot.py
@@ -4,9 +4,9 @@ import asyncio
 import inspect
 import logging
 import math
+import time
 import types
 from datetime import datetime
-import time
 from typing import TYPE_CHECKING, cast
 
 import aiohttp
--- a/ballsdex/core/bot.py
+++ b/ballsdex/core/bot.py
@@ -6,6 +6,7 @@ import logging
 import math
 import types
 from datetime import datetime
+import time
 from typing import TYPE_CHECKING, cast
 
 import aiohttp
@@ -419,7 +420,7 @@ class BallsDexBot(commands.AutoShardedBot):
             if isinstance(error, app_commands.CommandOnCooldown):
                 await send(
                     "This command is on cooldown. Please retry "
-                    f"in {math.ceil(error.retry_after)} seconds."
+                    f"in <t:{time.time() + math.ceil(error.retry_after)}:R> seconds."
                 )
                 return
             await send("You are not allowed to use that command.")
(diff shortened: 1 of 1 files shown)


flaree:
I don't see the point in adding this to be honest, it's just adding more computation to something that is just a simple message saying X can be ran in Y seconds.



@laggron42 thoughts?

Buggem:
> I don't see the point in adding this to be honest, it's just adding more computation to something that is just a simple message saying X can be ran in Y seconds.
> 
> @laggron42 thoughts?

It doesn't tick. My pea-brain struggles to keep count of when its ready.

ballsdex/core/bot.py
```diff
@@ -419,7 +420,7 @@ async def send(content: str):
             if isinstance(error, app_commands.CommandOnCooldown):
                 await send(
                     "This command is on cooldown. Please retry "
-                    f"in {math.ceil(error.retry_after)} seconds."
+                    f"in <t:{time.time() + math.ceil(error.retry_after)}:R> seconds."
```
laggron42, line 423: A few errors in the format, just had to group everything in the `math.ceil` and remove the words around it (Discord already writes "in x seconds").
image


```suggestion
                    f"<t:{math.ceil(time.time() + error.retry_after)}:R>."
```

laggron42 (commented):
I think it's a good addition! It lets you see the countdown before having to try again, and I find that very cool. Just a little thing to change and we're good to go.

ballsdex/core/bot.py
Buggem: Commited :D

flaree:
Thanks for addressing the changes, much appreciated!

Just checking, if you're aware on how to use pre-commit to run the checks so the build won't fail.

No worries if not! We can do it ourselves, once I'm back from vacation I can do it 😁

Buggem:
> Thanks for addressing the changes, much appreciated!
> 
> Just checking, if you're aware on how to use pre-commit to run the checks so the build won't fail.
> 
> No worries if not! We can do it ourselves, once I'm back from vacation I can do it 😁

No I don't sorry, could you help (once you get back?)

laggron42:
There is documentation in the [contribution guide](https://github.com/Ballsdex-Team/BallsDex-DiscordBot/blob/master/CONTRIBUTING.md#coding-style) (the `pre-commit` command must be ran after running `poetry shell`, otherwise it may not exist).

I did the fix for you in the meantime, just to keep in mind for later!

PR #285, opened 2024-06-19
adding `special` and `shiny` arg to `/balls favourite`

imtherealF1:
it was missing from there, so why not add it, helps make everything faster for users with many countryballs.

### DIFF
--- a/ballsdex/packages/balls/cog.py
+++ b/ballsdex/packages/balls/cog.py
@@ -417,8 +417,8 @@ class Balls(commands.GroupCog, group_name=settings.players_group_cog_name):
 
     @app_commands.command()
     async def favorite(
-        self, 
-        interaction: discord.Interaction, 
+        self,
+        interaction: discord.Interaction,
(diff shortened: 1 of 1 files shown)


ltzmax (commented):
While testing this: This didn't really show any special or and shiny in the selecting. It still refer to every balls that are not special or and shiny.

imtherealF1:
interesting, i tested this on my self host just now and it seems to work correctly, both shinies and specials. to be honest, i cant see where there would be a problem, as this pr is the same as what i did with balls give and trade remove a while back. if you see any problems or possible changes please tell me.

ltzmax:
> interesting, i tested this on my self host just now and it seems to work correctly, both shinies and specials. to be honest, i cant see where there would be a problem, as this pr is the same as what i did with balls give and trade remove a while back. if you see any problems or possible changes please tell me.

No dont worry, looks like discord was being discord. haha, works great!

ltzmax (approved):

Issue #293, opened 2024-06-25
Bot Crashing on Message event


TwixGamer00:
Hey!

I tried running the bot (in Docker) but I can't quite get it to work. I have some experience with discord.js but not really with discord.py. The bot starts without any problems, however, as soon as a message event is received the bot crashes. I enabled all 3 intents in the developer portal.

Im not quite wure what's causing it. Is it possibke that i have the wrong version of something?

Docker Logs:
````
2024-06-25 12:45:07 2024-06-25 10:45:07 INFO     ballsdex.core.bot Synced 1 admin commands for guild 1182615554897477682.
2024-06-25 12:45:17 2024-06-25 10:45:17 CRITICAL ballsdex Caught unhandled exception in <Task finished name='discord.py: on_message' coro=<Client._run_event() done, defined at /usr/local/lib/python3.11/site-packages/discord/client.py:433> exception=TypeError('sequence item 0: expected str instance, Message found')>:
2024-06-25 12:45:17 Task exception was never retrieved
2024-06-25 12:45:17 Traceback (most recent call last):
2024-06-25 12:45:17   File "/usr/local/lib/python3.11/site-packages/discord/ext/commands/bot.py", line 1299, in get_context
2024-06-25 12:45:17     if origin.content.startswith(tuple(prefix)):
2024-06-25 12:45:17        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-06-25 12:45:17 TypeError: tuple for startswith must only contain str, not NoneType
2024-06-25 12:45:17 
2024-06-25 12:45:17 During handling of the above exception, another exception occurred:
2024-06-25 12:45:17 
2024-06-25 12:45:17 Traceback (most recent call last):
2024-06-25 12:45:17   File "/usr/local/lib/python3.11/site-packages/discord/client.py", line 441, in _run_event
2024-06-25 12:45:17     await coro(*args, **kwargs)
2024-06-25 12:45:17   File "/usr/local/lib/python3.11/site-packages/discord/ext/commands/bot.py", line 1395, in on_message
2024-06-25 12:45:17     await self.process_commands(message)
2024-06-25 12:45:17   File "/usr/local/lib/python3.11/site-packages/discord/ext/commands/bot.py", line 1390, in process_commands
2024-06-25 12:45:17     ctx = await self.get_context(message)
[...]

PR #294, opened 2024-06-25
Handle every text-prefix as a string

MaxxMXSZ:
related to issue in #293 this will not happen in the further!


ballsdex/settings.py
```diff
@@ -93,7 +93,7 @@ def read_settings(path: "Path"):
     settings.bot_token = content["discord-token"]
     settings.gateway_url = content.get("gateway-url")
     settings.shard_count = content.get("shard-count")
-    settings.prefix = content["text-prefix"]
+    settings.prefix = str(content.get("text-prefix"))
```
laggron42, line 96: Let's add a default there
```suggestion
    settings.prefix = str(content.get("text-prefix", "b."))
```

PR #297, opened 2024-06-27
slight issue fix with admin info guild

imtherealF1:
when i was testing things with my self host, i made a new server, with no pfp, and when i ran the cmd admin info guild it failed, this fixes it!


flaree (approved):

flaree:
Thanks for the PR :)

PR #298, opened 2024-06-27
in many descriptions, changing "ball" to "countryball"

imtherealF1:
this of course is so that in other dexes, the "countryball" will get translated as {settings.collectible_name}.


flaree (approved):

flaree:
Thanks again for the PR

PR #299, opened 2024-06-28
Add more customizability specifically phrases

MaxxMXSZ:


ballsdex/settings.py
```diff
@@ -48,6 +48,18 @@ class Settings:
         List of roles that have full access to the /admin command
     admin_role_ids: list[int]
         List of roles that have partial access to the /admin command (only blacklist and guilds)
+    wild_phrase: str
+        The phrase used when a collectible spawns
+    wrong_name: str
+        The phrase used when a user enters the wrong name for that collectible
+    you_caught: str
+        The phrase used when a user catch a ball
+    new_comp: str = ""
```
flaree, line 57: Could there be a better name for this, seems to be confusing without digging for the docstring

ballsdex/packages/countryballs/components.py
```diff
@@ -50,10 +50,13 @@ async def on_error(self, interaction: discord.Interaction, error: Exception, /)
     async def on_submit(self, interaction: discord.Interaction["BallsDexBot"]):
         # TODO: use lock
         if self.ball.catched:
-            await interaction.response.send_message(
-                f"{interaction.user.mention} I was caught already!"
-            )
-            return
+            if settings.mention_user:
+                message = f"{interaction.user.mention} {settings.caught_already_phrase}"
```
flaree, line 54: small nitpick, doesn't make much of a difference - you can use discord.AllowedMentions so it doesn't actually ping instead of having two separate messages

PR #300, opened 2024-06-28
Small fix when comparing names

TheHappyBee:
Dear Ballsdex developer team,
Is it possible that possible names for countryballs can sometimes not match due to the fact that input from the user is lowered and the text field in the Balls model -> possible_names can accidently contain uppercase letters. Shouldn't something like this be done to prevent it from mismatching or am I mistaken and something is already done to avoid this. I have opened a pr for this reason.
Thanks,
HappyBee

### DIFF
--- a/ballsdex/packages/countryballs/components.py
+++ b/ballsdex/packages/countryballs/components.py
@@ -55,7 +55,7 @@ class CountryballNamePrompt(Modal, title=f"Catch this {settings.collectible_name
             )
             return
         if self.ball.model.catch_names:
-            possible_names = (self.ball.name.lower(), *self.ball.model.catch_names.split(";"))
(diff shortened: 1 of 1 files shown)


TheHappyBee:
Hello I have fixed black formatting issue. Thought I could get away with readability after writing only a little bit of code.

laggron42:
Hey, the catch names should always be lowered and stripped by this function: https://github.com/Ballsdex-Team/BallsDex-DiscordBot/blob/master/ballsdex/core/models.py#L26-L36

You shouldn't have encountered a situation where those catch names aren't pre-lowered (unless you put them using the SQL console), is this an issue you have actually encountered?

TheHappyBee:
ok thank you so much for the insight, I missed that. I will close the pull request shortly

ManzorroGithub:
Hey I'm pretty new in here, why do countryball's abilities not work?

PR #316, opened 2024-07-17
[admin] Add event create command

Valen7440:
A little help for admins that can't access the panel for some reason


laggron42:
Why closing? That seemed interesting

Valen7440:
Because it's not working properly

PR #325, opened 2024-08-12
[policies] add mention policy

flaree:


migrations/models/32_20240810180151_update.sql
```diff
@@ -0,0 +1,32 @@
+-- upgrade --
```
Kowlin, line 1: These columns seem off, they mention ``coins``, something not reflected in ``master`` yet. Could maybe do with a new SQL migration file?

Kowlin (commented):
LGTM, Just the migration file that needs addressing.

ltzmax (approved):

PR #327, opened 2024-08-12
[admin] Add admin blacklist history

flaree:
Also adds moderator ID to the models instead of appending it in the reason (yucky).

Add reasons for unlisting


ballsdex/packages/admin/cog.py
```diff
@@ -743,21 +755,59 @@ async def blacklist_info(
...
                     f"`{user}` (`{user.id}`) is currently blacklisted (date unknown)"
                     " for the following reason:\n"
-                    f"{blacklisted.reason}",
+                    f"{blacklisted.reason}\n{moderator_msg}",
                     ephemeral=True,
                 )
 
+    @blacklist.command(name="history")
+    @app_commands.checks.has_any_role(*settings.root_role_ids, *settings.admin_role_ids)
+    async def blacklist_history(self, interaction: discord.Interaction, id: str):
```
ltzmax, line 782: maybe have it as user_id instead?, id feels a bit confusions

ltzmax (commented):
Over all seems fine, just the question in the review and this should be good.
Also unblacklist reason are not posted to logs in discord channel, not sure if we want that?

ltzmax (approved):

PR #337, opened 2024-08-18
pyright version bumps

imtherealF1:


imtherealF1:
imma edit this when the other prs get looked at

imtherealF1:
due to the version bump being big, i need to fix a lot of errors that pyright gives, so should be many commits.

imtherealF1:
there are way too many conflicts on this, i will make a new branch

PR #338, opened 2024-08-19
adding primary key "id" to distinct, bug fix

imtherealF1:
im not sure if this solves the issue, but it should


imtherealF1:
it doesnt work, closing

PR #339, opened 2024-08-19
adding poetry lock --no-update to pre commit

imtherealF1:
the reason i think this is needed is because when someone bumps the version of a dependency, they will need to run poetry lock --no-update to write a new lock file, then copy that and paste it into the lock file manually. by adding this as a check, it will generate the lock file automatically.


Kowlin (approved):
LGTM

PR #348, opened 2024-08-21
Bumping the pyright version to 1.1.377

imtherealF1:


imtherealF1:
imma close this, pointless trying to bump the version now

PR #350, opened 2024-08-22
adding interaction check to config embed, making channel optional, and some logic changes

imtherealF1:


ballsdex/packages/config/cog.py
```diff
@@ -40,36 +40,53 @@ def __init__(self, bot: "BallsDexBot"):
...
-        if not channel.permissions_for(guild.me).read_messages:
-            await interaction.response.send_message(
-                f"I need the permission to read messages in {channel.mention}."
-            )
-            return
-        if not channel.permissions_for(guild.me).send_messages:
+
+        # Check if channel is not None before accessing its permissions
+        if channel:
+            if not channel.permissions_for(guild.me).read_messages:
```
flaree, line 62: https://discordpy.readthedocs.io/en/stable/interactions/api.html?highlight=has_permission#discord.app_commands.checks.has_permissions


These checks can be done in the decorator to save having repeated logic

ballsdex/packages/config/components.py
```diff
@@ -10,10 +10,18 @@ class AcceptTOSView(View):
     Button prompting the admin setting up the bot to accept the terms of service.
     """
 
-    def __init__(self, interaction: discord.Interaction, channel: discord.TextChannel):
+    def __init__(
+        self,
+        interaction: discord.Interaction,
+        channel: discord.TextChannel,
```
flaree, line 16: What's the need for the player object here? Just pass the discord.Member object instead directly from the invoked of the command

flaree (changes requested):
Couple changes to reduce repeated code and some other issues

ballsdex/packages/config/cog.py
imtherealF1: @flaree when testing @app_commands.checks.bot_has_permissions(read_messages=True, send_messages=True, embed_links=True), i realised that when the bot doesnt have the permissions required, it responds with "You are not allowed to use this command", which doesnt state what the issue is. should i keep that or should i change it

flaree: That's fine for now, i'll create a separate issue to handle permission checks correctly

ballsdex/packages/config/cog.py
```diff
@@ -36,40 +36,56 @@ def __init__(self, bot: "BallsDexBot"):
...
+    )
     async def channel(
         self,
         interaction: discord.Interaction,
-        channel: discord.TextChannel,
+        channel: Optional[discord.TextChannel] = None,
     ):
         """
-        Set or change the channel where countryballs will spawn.
+        The new text channel to set.
```
flaree, line 50: I think the previous description is more apt, there's no context as to what you want to set here

ballsdex/packages/config/cog.py
```diff
@@ -36,40 +36,56 @@ def __init__(self, bot: "BallsDexBot"):
...
+        ----------
+        channel: discord.TextChannel
+            The channel you want to set, current one if not specified.
         """
-        guild = cast(discord.Guild, interaction.guild)  # guild-only command
         user = cast(discord.Member, interaction.user)
+
+        if channel is None:
+            if isinstance(interaction.channel, discord.TextChannel):
+                channel = cast(discord.TextChannel, interaction.channel)
```
flaree, line 61: What's the use of this casting, your isinstance check is True meaning it already is a TextChannel

ballsdex/packages/config/cog.py
```diff
@@ -36,40 +36,56 @@ def __init__(self, bot: "BallsDexBot"):
...
-            )
-            return
-        if not channel.permissions_for(guild.me).send_messages:
-            await interaction.response.send_message(
-                f"I need the permission to send messages in {channel.mention}."
-            )
-            return
-        if not channel.permissions_for(guild.me).embed_links:
+
+        if channel and isinstance(channel, discord.TextChannel):
```
flaree, line 74: May be better to separate the handling of the instance check out instant of having it in this if and elif

ballsdex/packages/config/components.py
```diff
@@ -29,40 +39,46 @@ def __init__(self, interaction: discord.Interaction, channel: discord.TextChanne
...
-        except discord.HTTPException:
-            pass
+        if self.message:
+            for item in self.children:
+                if isinstance(item, discord.ui.Button):
+                    item.disabled = True
+            try:
+                await self.message.edit(view=self)
+            except discord.HTTPException:
+                pass
```
ltzmax, line 84: the `self.stop()` is not really needed
```suggestion
    async def on_timeout(self) -> None:
        if self.message:
            for item in self.children:
                item: discord.ui.Button
                item.disabled = True
            try:
                await self.message.edit(view=self)
            except discord.HTTPException:
                pass
```

flaree (approved):

PR #357, opened 2024-08-29
adding dimensions to the collection card label in admin/resources

imtherealF1:


flaree (commented):
could this be expanded and also have the correct dimensions for wildcard art too?

flaree (approved):

PR #360, opened 2024-08-29
handling permission checks correctly with details

### LINKED ISSUE #351: Handle permission errors correctly
The generic the bot doesn't have permission to do something or you can not run this command checks need to be expanded upon and elaborated further.

bot has permissions check should return the missing permissions, some other stuff should fail more silently

imtherealF1:
closes #351


flaree (changes requested):
line 368: 

theres a return early for a CheckFailure, meaning for regular commands your code isnt ran

flaree (approved):

PR #368, opened 2024-08-30
[config] /config silent, prevents notification spam

### LINKED ISSUE #68: Config option for silent ball spawning
Having a config option like `/config silent` which makes it so ball spawns and the resulting messages from guesses all get sent with the `silent` arg set to `True` would be useful so that it prevents notification spam in channels where users typically have them enabled.

I am willing to submit a PR for the feature if this is approved.

imtherealF1:
every response from guessing messages except catch message and spawn is ephemeral.

closes #68

### DIFF
--- a/ballsdex/packages/config/cog.py
+++ b/ballsdex/packages/config/cog.py
@@ -58,6 +58,7 @@ class Config(commands.GroupCog):
         self,
         interaction: discord.Interaction,
         channel: Optional[discord.TextChannel] = None,
+        silent: bool = False,
     ):
         """
         Set or change the channel where countryballs will spawn.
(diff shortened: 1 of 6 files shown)


ballsdex/packages/config/cog.py
```diff
@@ -66,14 +78,53 @@ async def channel(
...
-        view = AcceptTOSView(interaction, channel, user)
+        view = AcceptTOSView(interaction, channel, user, silent=False)
         message = await channel.send(embed=activation_embed, view=view)
         view.message = message
 
         await interaction.response.send_message(
             f"The activation embed has been sent in {channel.mention}.", ephemeral=True
         )
 
+    @app_commands.command()
```
flaree, line 89: Im not a fan of having a seperate command, I would prefer adding a boolean option to the /config channel command

flaree (approved):

PR #374, opened 2024-08-31
updating a lot of dependencies

imtherealF1:


Kowlin:
TODO: Check for dependency removals and other stuff.

pyproject.toml
```diff
@@ -9,34 +9,34 @@ license = "MIT"
...
 
 # misc
-rich = "^13.7.1"
+rich = "^13.8.0"
 python-dateutil = "^2.9.0"
 Pillow = "^10.4.0"
-aerich = "^0.6.3"
-pyyaml = "^6.0.1"
-cachetools = "^5.3.3"
+aerich = "^0.7.2"
```
laggron42, line 30: Nope nope nope not that!!

Aerich introduced breaking changes in 0.7.X, using Python files instead of SQL files for migrations, with no easy way to convert our existing migrations. This must absolutely stay pinned on 0.6.X.

laggron42 (changes requested):

laggron42 (approved):

PR #375, opened 2024-09-01
[balls] changing syntax for stats through balls list

imtherealF1:
removing the caught on, makes it look better

### DIFF
--- a/ballsdex/packages/balls/countryballs_paginator.py
+++ b/ballsdex/packages/balls/countryballs_paginator.py
@@ -41,7 +41,7 @@ class CountryballsSelector(Pages):
                     description=(
                         f"ATK: {ball.attack}({ball.attack_bonus:+d}%) "
                         f"• HP: {ball.health}({ball.health_bonus:+d}%) • "
-                        f"Caught on {ball.catch_date.strftime('%d/%m/%y %H:%M')}"
+                        f"{ball.catch_date.strftime('%d/%m/%y %H:%M')}"
                     ),
                     emoji=emoji,
                     value=f"{ball.pk}",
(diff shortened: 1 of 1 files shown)


ballsdex/packages/balls/countryballs_paginator.py
```diff
@@ -41,7 +41,7 @@ def set_options(self, balls: List[BallInstance]):
                     description=(
                         f"ATK: {ball.attack}({ball.attack_bonus:+d}%) "
                         f"• HP: {ball.health}({ball.health_bonus:+d}%) • "
-                        f"Caught on {ball.catch_date.strftime('%d/%m/%y %H:%M')}"
+                        f"{ball.catch_date.strftime('%d/%m/%y %H:%M')}"
```
Kowlin, line 44: Could possible also be shortened to just ``%y/%m/%d`` (not ``%d/%m/%y``) by omitting the hour and minute marks.

imtherealF1: it is a bit confusing like that tbh, having d/m/y seems better

Kowlin: Allow me to raise you the issue of being European 😛, Europe has a ``dd/mm/yyyy``, the US has a format of ``mm/dd/yyyy``, Is this stupid and confusing?! YES! Has this caused endless arguments, Absolutely. Which caused the spawn of [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601), which suggests ``yyyy/mm/dd`` as a middle ground. Is this needlessly confusing? yes!~ But because we cannot use relative timestamps in lists. Its picking one of the worst worlds, or... the middle ground of ISO 8601.

ltzmax (approved):

PR #376, opened 2024-09-01
[poetry] Bump tortoise orm (0.21.4 to 0.21.6)

Valen7440:


flaree:
closing, i see no need to bump dependencies as of yet + we have a big dependency pr open currently

PR #380, opened 2024-09-03
[trade] adding select all pages button in trade bulk add

imtherealF1:


flaree:
Sorry, I'm going to be declining this.

I'm not a fan of a massive bulk operation like this as people will easily trade stuff they don't mean to, plus makes the menu more bulkier. The selecting of a single page will be the max it'll be

PR #381, opened 2024-09-03
[trade] changing "this interaction ends in 30 mins" to a discord timestamp

imtherealF1:


ballsdex/packages/trade/menu.py
```diff
@@ -155,6 +156,7 @@ def __init__(
         self.task: asyncio.Task | None = None
         self.current_view: TradeView | ConfirmView = TradeView(self)
         self.message: discord.Message
+        self.end_time = math.ceil((datetime.now(timezone.utc) + timedelta(minutes=30)).timestamp())
```
ltzmax, line 159: You can instead use this over having extra to import. It's a part of [discord.py](https://discordpy.readthedocs.io/en/stable/api.html#discord.utils.format_dt)
```suggestion
        self.end_time = discord.utils.format_dt(datetime.now(timezone.utc) + timedelta(minutes=30))
```

ltzmax (changes requested):

ballsdex/packages/trade/menu.py
```diff
@@ -175,7 +177,7 @@ def _generate_embed(self):
             f"using the {add_command} and {remove_command} commands.\n"
             "Once you're finished, click the lock button below to confirm your proposal.\n"
             "You can also lock with nothing if you're receiving a gift.\n\n"
-            "*You have 30 minutes before this interaction ends.*\n\n"
+            f"*This interaction ends {self.end_time}.*\n\n"
```
flaree, line 180: Is there a need to use a variable here since it's only being called once?

flaree (approved):

PR #382, opened 2024-09-03
add stats (no stats bonus) in description

Valen7440:


imtherealF1:
personally, i dont like how the design is, it has way too many information, and since this will mostly be used in autocomplete for commands such as /balls give, /balls favourite, i dont think that a lot of info is indeed required. maybe if the syntax is improved it will look better.

flaree:
On a separate note, please raise an issue in regarding to changing something design wise. Just saves time if it was to be denied in the issue instead of spending time implementing

Valen7440:
Ok

PR #383, opened 2024-09-04
[settings] making the biggest defense/attack stat bonus configurable

imtherealF1:


json-config-ref.json
```diff
@@ -3,7 +3,20 @@
     "title": "Ballsdex configuration",
     "description": "Core settings for the Ballsdex Discord bot",
     "type": "object",
-    "required": ["discord-token", "text-prefix", "about", "collectible-name", "bot-name", "players-group-cog-name", "admin-command", "prometheus", "owners"],
+    "required": [
+        "discord-token",
+        "text-prefix",
+        "about",
+        "collectible-name",
+        "max-favorites",
```
flaree, line 11: Personally I don't think it should be required, as in the config settings you're defaulting it to 20 anyhow

ballsdex/settings.py
```diff
@@ -34,6 +34,10 @@ class Settings:
         Set the name of the base command of the "players" cog, /balls by default
     max_favorites:
         Set the maximum amount of favorited countryballs a user can have, 50 by default.
+    max_attack_bonus:
+        Set the biggest/smallest attack bonus that a spawned countryball can have.
+    max_defense_bonus:
```
flaree, line 39: Can it be renamed to max_health_bonus instead to reflect that it's actually health and not defence

flaree (changes requested):

flaree (approved):

PR #387, opened 2024-09-06
[Balls] move translations to a new field for easier management

flaree:
Seperate translations from catch names

Allows for easier updating of translations without affecting alt catch names from countries with accents etc


ltzmax (approved):
LGTM

PR #388, opened 2024-09-06
[Models] Add soft delete field 

flaree:
Adds a deleted field to BallInstances

Allows us to mark a ball as deleted without it actually getting deleted and affecting other models such as Trade History etc.

### DIFF
--- a/ballsdex/packages/admin/cog.py
+++ b/ballsdex/packages/admin/cog.py
@@ -1165,13 +1165,17 @@ class Admin(commands.GroupCog):
         if not view.value:
             return
         if percentage:
-            balls = await BallInstance.filter(player=player)
+            balls = await BallInstance.filter(
+                player=player
+            )  # TODO: Do we delete perm or soft delete here?
             to_delete = random.sample(balls, int(len(balls) * (percentage / 100)))
             for ball in to_delete:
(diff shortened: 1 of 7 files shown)


ballsdex/packages/admin/cog.py
```diff
@@ -1165,13 +1165,17 @@ async def balls_reset(
         if not view.value:
             return
         if percentage:
-            balls = await BallInstance.filter(player=player)
+            balls = await BallInstance.filter(
```
flaree, line 1168: @laggron42 @Kowlin input needed here, in `/admin balls reset` should we permanently delete or soft delete

laggron42: I suggest soft delete with an argument to hard reset (defaults to soft)

laggron42 (commented):
I think you forgot to commit the new column in `models.py`, not sure how pyright validated this

flaree:
> I think you forgot to commit the new column in `models.py`, not sure how pyright validated this

image
?

laggron42:
> > I think you forgot to commit the new column in `models.py`, not sure how pyright validated this
> 
> image ?

My bad, Github wasn't showing me the full changes

PR #395, opened 2024-09-12
[trade] checking if the user is in the server before creating the trade

imtherealF1:


imtherealF1:
realised this would only work for bots with member intent, closing

PR #405, opened 2024-09-13
player2 blacklist check in /friend add + handling for mentions policy

imtherealF1:


ballsdex/packages/players/cog.py
```diff
@@ -179,6 +179,11 @@ async def friend_add(self, interaction: discord.Interaction, user: discord.User)
         if user.bot:
             await interaction.response.send_message("You cannot add a bot.", ephemeral=True)
             return
+        if player2.discord_id in self.bot.blacklist:
+            await interaction.response.send_message(
+                "You cannot add a blacklisted user as a friend.", ephemeral=True
```
Kowlin, line 184: This string should be changed to something less detail oriented IMO, It feels like it can be used for blacklist scouting.

flaree: We have the same responses elsewhere, so its fine

imtherealF1: i dont think so, you can already do that with /balls give nevertheless

flaree (approved):

PR #412, opened 2024-09-15
Change settings.collectible_name to settings.plural_collectible_name

Valen7440:


imtherealF1:
you're not really doing anything? you're not using plural collectible name, you're just removing the s, it would result in the bot saying countryball only

PR #416, opened 2024-09-15
[balls] grammar fix

imtherealF1:


flaree:
This will add an extra space if no special is passed

ltzmax (approved):
package

PR #418, opened 2024-09-16
[config] change disable to toggle

### LINKED ISSUE #415: Rename /config disable to /config toggle
Renaming /config disable to /config toggle.

**Additional context**
This makes more sense as upon running /config disable and disabling countryball spawns, it prompts you to run the command again to re-enable it, and this is effectively just a toggle for toggling countryball spawns, so that's the main reason why.

flaree:
closes #415


ltzmax (commented):
Maybe we can remove it from being optional to use `enabled: `? Right now based on the  wording, it says to redo the command to re enable, which might be confusing for our user base?

R641c8

userandaname:
> Maybe we can remove it from being optional to use `enabled: `? Right now based on the wording, it says to redo the command to re enable, which might be confusing for our user base?
> 
> R641c8

i'm not the one to really say here, but i'd say that it'd make more sense to make it optional in the case that users are already used to /config enable, and so bot-hosters forking BallsDex may want to change that option based on their user-base?

laggron42:
> i'm not the one to really say here, but i'd say that it'd make more sense to make it optional in the case that users are already used to /config enable, and so bot-hosters forking BallsDex may want to change that option based on their user-base?

It's going to be super hard to have two conditional blocks of code, if self hosters don't like the way we're going they're free to fork/stop updating, we focus on our own user base. I try to consider the selfhosters when adding or changing stuff, but having this sort of switch is really overcomplicated for nothing.

PR #419, opened 2024-09-16
[trade] adding a check for locked for trade + favorited balls for bulk add

imtherealF1:


ltzmax (changes requested):
This makes trade bulk add failing. it seems that it cannot be done this way, maybe do similar to trade add.

imtherealF1:
ill do that ye

ltzmax (approved):
📦

PR #421, opened 2024-09-17
[player] checking if there is an active friend request before sending another one

### LINKED ISSUE #424: Friend list showing same friend twice
**Describe the bug**
Friend list is showing same friend twice. My bot user found this and I think its bug
**To Reproduce**
1. use command /player friend list

**Expected behavior**
I think it should show same user only one time

**Screenshots**
image

**Version:**
 - 2.20.0

**Additional context**

imtherealF1:
closes #424

### DIFF
--- a/ballsdex/packages/players/cog.py
+++ b/ballsdex/packages/players/cog.py
@@ -194,6 +194,11 @@ class Player(commands.GroupCog):
     async def friend_add(self, interaction: discord.Interaction, user: discord.User):
         """
         Add another user as a friend.
+        
+        Parameters
+        ----------
+        user: discord.User
+            The user you want to add as a friend.
         """
         player1, _ = await PlayerModel.get_or_create(discord_id=interaction.user.id)
         player2, _ = await PlayerModel.get_or_create(discord_id=user.id)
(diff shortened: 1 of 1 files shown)


ltzmax (commented):
I would just use `max_concurrency` check instead of doing it this way.

imtherealF1:
> I would just use `max_concurrency` check instead of doing it this way.

issue is, its not available for app_commands, i already tried it, its only for Command, and its something that cant be implemented

ltzmax:
> > I would just use `max_concurrency` check instead of doing it this way.
> 
> issue is, its not available for app_commands, i already tried it, its only for Command, and its something that cant be implemented

Oh right, i forgot that has to be coded to have as its own decoder.

ltzmax:
> even if I would code that, the issue is that it wouldn't allow you to send any friend requests to any users after the max concurrency was reached, with this code you can send to any user that you haven't sent an active friend request

Not exactly. you gotta pass it into a class then build the decoder and handlers, i have codes that works with it but dw, we not gonna do that. 📦

ltzmax:
To me this looks pretty much fine and works, there's been no complains when we added this on a bigger bot than mine (about 800+ server sized bot), but i am also unsure if this is the best way without causing issues. 
I'll leave this up to review for @laggron42, @Kowlin or @MaxxMXSZ.

PR #432, opened 2024-09-30
[info] new /tutorial command

### LINKED ISSUE #377: A command that shows a tutorial for BallsDex!
The command name would be /tutorial, in the info package. What it would do is it would send an embed, possibly with pagination too, that says how to play, what you can do in the bot, general info about the bot's gameplay, but it would also say where you can suggest a feature, or report a possible bug, where to find support, and the footer can be a link to the bot TOS.

**Additional context**
I'm suggesting this because of the increasing amount of new users, and a lot of them aren't sure on how to start, what to do, how to use the bot. A command like this would help. /help gives every command with its description, so if we get a /tutorial command it would need to be different.

I can make a pr for this if it gets accepted.

imtherealF1:
closes #377


ballsdex/packages/info/cog.py
```diff
@@ -167,3 +169,65 @@ async def help(self, interaction: discord.Interaction):
...
+            ),
+            inline=False,
+        )
+
+        embed.add_field(
+            name="How can I configure the bot?",
+            value=(
+                f"To enable the spawning of {settings.plural_collectible_name}, you "
+                "need to configure the bot. To do that, you need to run the command "
+                f"{config_channel}. You must have the `Manage Guild` permission in that server."
```
ltzmax, line 200: I dont think we need to mention in the current server, so it'd be better this way instead.
```suggestion
                f"{config_channel}. You must have the ``Manage Guild`` "
                "permission to use this command."
```

ltzmax (changes requested):
i think this can be used on ballsdex.com/tutorial instead of making something like this but that's me.

imtherealF1:
it can, but since the source code for the website isn't here i just decided to do it like this for now

ballsdex/packages/info/cog.py
```diff
@@ -197,7 +197,7 @@ async def tutorial(self, interaction: discord.Interaction):
             value=(
                 f"To enable the spawning of {settings.plural_collectible_name}, you "
                 "need to configure the bot. To do that, you need to run the command "
-                f"{config_channel}. You must have the `Manage Guild` permission in that server."
+                f"{config_channel}. You must have the `Manage Guild` permission."
```
ltzmax, line 200: Still should say `to use this command` after permission, its more clear, because of our community base.

imtherealF1:
closing this too, ill reopen a new branch

PR #439, opened 2024-10-11
[trade] bug fix, plural check and more precise wording in /view and /begin

imtherealF1:


ltzmax (approved):
Works fine on my end!🛎️

PR #440, opened 2024-10-12
[balls] correcting the wording on is_blocked checks

imtherealF1:


ballsdex/packages/balls/cog.py
```diff
@@ -174,7 +174,13 @@ async def list(
         blocked = await player.is_blocked(interaction_player)
         if blocked and not is_staff(interaction):
             await interaction.followup.send(
-                "You cannot view the list of a user that has you blocked.", ephemeral=True
+                "You cannot view the list of a user that has blocked you.", ephemeral=True
+            )
+            return
+        blocked2 = await interaction_player.is_blocked(player)
```
flaree, line 180: Personally, I think if you blocked a user you can still view their inventory and only they cannot view yours

That's just my two cents

imtherealF1: i think that if you block a user you should avoid all interactions with them, but that's just me personally too, id like to hear another opinion too

ltzmax: Yeah, i also think you should be able to view their inventory until they block you as well. There's no reason to not being able to.

laggron42: I also think you should be able to see the inventory of a user you blocked, otherwise it discourages blocking for valid reasons.

PR #441, opened 2024-10-12
[trade] changing dictionary to ttlcache

### LINKED ISSUE #160: Trade not cleaning up
Various issues may occur leading to a trade not being cleaned up, thus resulting in a memory leak.

The trade dictionary should probably be changed to a TTLCache or introduce or own handling of it for finer control.

imtherealF1:
closes #160

### DIFF
--- a/ballsdex/packages/trade/cog.py
+++ b/ballsdex/packages/trade/cog.py
@@ -1,4 +1,5 @@
 import datetime
+from cachetools import TTLCache
 from collections import defaultdict
 from typing import TYPE_CHECKING, Optional, cast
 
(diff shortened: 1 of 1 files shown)


ballsdex/packages/trade/cog.py
```diff
@@ -35,7 +36,7 @@ class Trade(commands.GroupCog):
 
     def __init__(self, bot: "BallsDexBot"):
         self.bot = bot
-        self.trades: dict[int, dict[int, list[TradeMenu]]] = defaultdict(lambda: defaultdict(list))
+        self.trades = TTLCache(maxsize=999999, ttl=1800)
```
laggron42, line 39: Needs typing
```suggestion
        self.trades: TTLCache[int, dict[int, list[TradeMenu]]] = TTLCache(maxsize=999999, ttl=1800)
```

laggron42 (approved):

PR #442, opened 2024-10-12
Make admin reset "a little easier"

MaxxMXSZ:


ballsdex/packages/admin/cog.py
```diff
@@ -1148,15 +1148,17 @@ async def balls_reset(
                 "The user you gave does not exist.", ephemeral=True
             )
             return
-        if percentage and not 0 < percentage < 100:
+        if percentage and not 0 < percentage <= 101:
```
laggron42, line 1151: You're allowing `101` as a value there
```suggestion
        if percentage and not 0 < percentage <= 100:
```

laggron42 (commented):
You should add a check at line 1180 to ensure the percentage is below 100. While it would theorically work, better avoid unecessary code and logic, and skip straight to the "delete everything" case.

MaxxMXSZ:
i will check later not home rn

PR #450, opened 2024-10-23
[player] bug fix, not allowing users to be blocked during active friend req

imtherealF1:


ltzmax (approved):
After testing it for sometimes today it seems fine and code look fine!

Though i was a little nervous of the wording here, but i believe it should be fine to understand!

PR #451, opened 2024-10-23
[trade] adding a mandatory 5s cooldown between lock and accept

### LINKED ISSUE #443: [trade] Add mandatory cooldown in between lock and confirm
Add a 15-30 second mandatory cooldown in between the actions of locking and confirming

This will give people time to double check their trade and just blindly confirm

imtherealF1:
resolves #443


laggron42 (commented):
Hey, thanks for your work!

I'm pretty confused by this PR. You're not handling the fact that there are two users in a trade, so when the second user will lock, the timer will reset, is this an intended behavior?

I also think 15 seconds is a bit high, maybe something like 5 seconds? I also think it would be much more intuitive if the accept button is greyed out instead, then enabled after 5 (or 15) seconds.

imtherealF1:
> I'm pretty confused by this PR. You're not handling the fact that there are two users in a trade, so when the second user will lock, the timer will reset, is this an intended behavior?

Yes, as the intention is to set the cooldown at the start of the accept phase (when the second user locks).

> I also think 15 seconds is a bit high, maybe something like 5 seconds? I also think it would be much more intuitive if the accept button is greyed out instead, then enabled after 5 (or 15) seconds.

I think 5 should be good, but since the whole purpose of this is to give time to double check the proposals, the purpose of the cooldown should be stated in the footer of the embed.

flaree:
I disagree with the greyed out button, it's much more confusing - can you change it showing a message like `This trade cannot be approved for X more seconds, please use this time to double check the items to prevent any unwanted trades.`

Also, as a compromise - keep it as 10s

flaree:
Thanks for addressing the changes!

flaree (approved):

PR #453, opened 2024-10-25
[docker] fixing the "backup_dir hasnt got permissions" error

imtherealF1:


docker-compose.yml
```diff
@@ -67,7 +67,7 @@ services:
     restart: always
     user: postgres:postgres
     volumes:
-      - /var/opt/pgbackups:/backups
```
ltzmax, line 70: What OS are you actually getting the issue from?

imtherealF1: "backup_dir redirects to a file without permissions" was popping up on my docker logs constantly both windows and linux, and this fixed it

ltzmax: Well this has never popped up to me on linux at all, the best fix is to give the location permission.

laggron42: It's an issue for Windows, and it's going to make it easier to sort tbh

laggron42 (approved):

PR #460, opened 2024-11-11
Show the actual emoji instead of just the ID on the admin panel

cobaltt7:


laggron42 (approved):
Thank you, that's a great addition!

PR #467, opened 2024-11-24
[admin panel] dark mode, and a toggle to switch from one mode to another

imtherealF1:
If anyone has any design suggestions, or maybe adding the js/css in a seperate folder, or literally any feedback, I would love to hear it.


flaree:
Thank you!

I just have one small comment, could you possibly look into the font colour for the boolean types. They still have like a grey-ish colour when on dark mode and can be a bit of a strain, otherwise lgtm

flaree (approved):

PR #470, opened 2024-11-24
beautify spawn.py

alexanderthiem:
I had some problem reading the code and I realised it is because the variable names are kinda meaningless, so I changed them, without changing the behaviour.

multiplier? - there are 2 distinct multipliers
delta what? - delta_time
amount? Thats the most meaningless name for a number? chance? it is a `threshhold`, a randomly determined threshold to be exact
amount and self.amount in the same function also confused me for a second

### DIFF
--- a/ballsdex/packages/countryballs/spawn.py
+++ b/ballsdex/packages/countryballs/spawn.py
@@ -86,7 +86,8 @@ class SpawnCooldown:
         A number starting at 0, incrementing with the messages until reaching `threshold`. At this
         point, a ball will be spawned next.
     threshold: int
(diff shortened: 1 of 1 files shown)


flaree:
Thanks for the PR and addressing the pipeline needs, those can be annoying when contributing to newer projects 😅 

I've requested a review from laggron as he's the main person behind the spawn algorithm and can more objectively review this PR

alexanderthiem:
Ok. Nice. I thought I would save time by not opening an IDE for that. How wrong I was.

laggron42 (approved):

PR #472, opened 2024-11-28
[settings] move packages to settings

Dotsian:
This change makes it easier for self-hosted owners to add custom packages (without having to modify the bot.py file).

### DIFF
--- a/ballsdex/core/bot.py
+++ b/ballsdex/core/bot.py
@@ -46,8 +46,6 @@ if TYPE_CHECKING:
 log = logging.getLogger("ballsdex.core.bot")
 http_counter = Histogram("discord_http_requests", "HTTP requests", ["key", "code"])
 
-PACKAGES = ["config", "players", "countryballs", "info", "admin", "trade", "balls"]
-
 
 def owner_check(ctx: commands.Context[BallsDexBot]):
     return ctx.bot.is_owner(ctx.author)
@@ -304,7 +302,7 @@ class BallsDexBot(commands.AutoShardedBot):
             await self.add_cog(Dev())
 
         loaded_packages = []
-        for package in PACKAGES:
+        for package in settings.packages:
             try:
                 await self.load_extension("ballsdex.packages." + package)
             except Exception:
@@ -327,7 +325,7 @@ class BallsDexBot(commands.AutoShardedBot):
         else:
             log.info("No command to sync.")
 
-        if "admin" in PACKAGES:
+        if "admin" in settings.packages:
             for guild_id in settings.admin_guild_ids:
                 guild = self.get_guild(guild_id)
                 if not guild:
--- a/ballsdex/core/bot.py
+++ b/ballsdex/core/bot.py
@@ -303,12 +303,14 @@ class BallsDexBot(commands.AutoShardedBot):
 
         loaded_packages = []
         for package in settings.packages:
+            package_name = package.split(".")[-1]
+
             try:
-                await self.load_extension("ballsdex.packages." + package)
+                await self.load_extension(package)
             except Exception:
-                log.error(f"Failed to load package {package}", exc_info=True)
+                log.error(f"Failed to load package {package_name}", exc_info=True)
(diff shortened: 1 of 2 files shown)


flaree:
While this is a great idea, I foresee issues in the future should we add more cogs to the base bot as it won't be propagated to everyone's configs downstream

laggron42:
Haha you got forward of me, this is actually a change I've planned to do but a little differently, by using full importable paths instead (like `ballsdex.packages.trade`). This would allow 3rd party cogs to exist in their own pip-installable repositories, thus allowing users to install additional cogs without having to manually drop files or edit code.

Do you think you could change the PR to reflect that? You'd also need to remove the prepended path [here](https://github.com/Ballsdex-Team/BallsDex-DiscordBot/blob/6a7e88b5132a571f8d36c83b20cf22df3e31a608/ballsdex/core/bot.py#L309).

> While this is a great idea, I foresee issues in the future should we add more cogs to the base bot as it won't be propagated to everyone's configs downstream

I can still think of ways to dynamically extend the list of cogs after this change. Not trivial, but doable, and the benefits are really nice.

laggron42:
Thanks for the change, I'll test it as soon as possible but it already looks good on a first glance

ballsdex/core/bot.py
```diff
@@ -304,13 +302,15 @@ async def on_ready(self):
             await self.add_cog(Dev())
 
         loaded_packages = []
-        for package in PACKAGES:
+        for package in settings.packages:
+            package_name = package.split(".")[-1]
```
laggron42, line 306: I like this better, it doesn't hide the full path for custom packages
```suggestion
            package_name = package.replace("ballsdex.packages.", "")
```

ballsdex/core/bot.py
```diff
@@ -327,7 +327,7 @@ async def on_ready(self):
         else:
             log.info("No command to sync.")
 
-        if "admin" in PACKAGES:
+        if "admin" in settings.packages:
```
laggron42, line 330: Doesn't sync without the full path
```suggestion
        if "ballsdex.packages.admin" in settings.packages:
```

ballsdex/settings.py
```diff
@@ -139,6 +143,8 @@ def read_settings(path: "Path"):
     settings.max_attack_bonus = content.get("max-attack-bonus", 20)
     settings.max_health_bonus = content.get("max-health-bonus", 20)
 
+    settings.packages = content["packages"] or []
```
laggron42, line 146: Because of odd design, the content of the yaml is read before update, so you need to handle the case where the list doesn't exist yet
```suggestion
    settings.packages = content.get("packages") or [
        "ballsdex.packages.admin",
        "ballsdex.packages.balls",
        "ballsdex.packages.config",
        "ballsdex.packages.countryballs",
        "ballsdex.packages.info",
        "ballsdex.packages.players",
        "ballsdex.packages.trade",
    ]
```

ballsdex/settings.py
```diff
@@ -247,6 +263,7 @@ def update_settings(path: "Path"):
     add_max_attack = "max-attack-bonus" not in content
     add_max_health = "max-health-bonus" not in content
     add_plural_collectible = "plural-collectible-name" not in content
+    add_packages = "packages" not in content
```
laggron42, line 266: Won't be detected otherwise as "ballsdex.packages.countryballs.spawn.SpawnManager" already exists in the default config
```suggestion
    add_packages = "packages:" not in content
```

laggron42 (changes requested):
I tested the PR and had to do the following changes to get it to work as intended. Apply as needed and I can merge

Dotsian:
> I tested the PR and had to do the following changes to get it to work as intended. Apply as needed and I can merge

Hi, I couldn't test the PR because I was out for the holidays. I just resolved all the changes requested 🙂

laggron42 (approved):
Tested and works for me, thanks a lot for the contribution!

PR #473, opened 2024-11-29
[player] fixing /block add

imtherealF1:


ballsdex/packages/players/cog.py
```diff
@@ -381,12 +381,13 @@ async def block_add(self, interaction: discord.Interaction, user: discord.User):
...
             await interaction.followup.send("You have already blocked this user.", ephemeral=True)
             return
-        if self.active_friend_requests[(player1.discord_id, player2.discord_id)]:
-            await interaction.followup.send(
-                "You cannot block a user to whom you have sent an active friend request.",
-                ephemeral=True,
-            )
-            return
+        if (player1.discord_id, player2.discord_id) in self.active_friend_requests:
+            if self.active_friend_requests[(player1.discord_id, player2.discord_id)]:
```
flaree, line 385:
```suggestion
if self.active_friend_requests.get((player1.discord_id, player2.discord_id)):
```
Can reduce it to just one line instead of having 2 checks

flaree (changes requested):

flaree (approved):

PR #479, opened 2024-12-13
[core] wording fix on bot start-up

imtherealF1:
on "synced x commands", changing the commands to cogs, as commands is misleading

### DIFF
--- a/ballsdex/core/bot.py
+++ b/ballsdex/core/bot.py
@@ -318,7 +318,7 @@ class BallsDexBot(commands.AutoShardedBot):
 
         synced_cogs = await self.tree.sync()
         grammar = "" if synced_cogs == 1 else "s"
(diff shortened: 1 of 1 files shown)


laggron42:
Cogs are not synced, that's indeed the number of commands you're seeing. 6 isn't even the number of cogs loaded:

- admin: 0 commands (they are synced per-guild so it doesn't count)
- balls: 1 group command
- config: 1 group command
- countryballs: 0 commands
- info: 2 commands
- players: 1 group command
- trade: 1 group command

That's 6 top-level commands, as intended and described in [the docs](https://discordpy.readthedocs.io/en/latest/interactions/api.html#discord.app_commands.CommandTree.sync)

imtherealF1:
ah nvm then

PR #483, opened 2024-12-20
Adding /Balls Duplicate

MaxxMXSZ:
Added `/Balls Duplicate`
If Countryballs is selected it shows top 50 of your most duped countryballs.
If Special is selected it shows top 5 of your most duped special countryballs.


MaxxMXSZ:
If you are not happy with the amount of specials shown. I can change it. But for now I think it's the perfect amount.

ballsdex/packages/balls/cog.py
```diff
@@ -660,3 +661,102 @@ async def count(
             f"You have {balls} {special_str}{shiny_str}"
             f"{country}{settings.collectible_name}{plural}{guild}."
         )
+
+    @app_commands.command()
```
laggron42, line 665: Command needs a cooldown, the queries running here are pretty intensive

ballsdex/packages/balls/cog.py
```diff
@@ -660,3 +661,102 @@ async def count(
...
+    @app_commands.choices(
+        type=[
+            app_commands.Choice(
+                name=f"{settings.collectible_name}", value=f"{settings.collectible_name}"
+            ),
+            app_commands.Choice(name="special", value="specials"),
+        ]
+    )
```
laggron42, line 673: I'm not a fan of decoratory-style parameters, can't you move that to an enum? The bot automatically transforms the word "countryball" into the configured collectible name

ballsdex/packages/balls/cog.py
```diff
@@ -660,3 +661,102 @@ async def count(
...
+
+        Parameters
+        ----------
+        type : app_commands.Choice[str]
+            Type of duplicate to check (Countryball or Special).
+        """
+        await interaction.response.defer(ephemeral=True)
+        user_id = interaction.user.id
+        player, _ = await Player.get_or_create(discord_id=user_id)
+        await player.fetch_related("balls", "balls__special")
```
laggron42, line 688: I don't think the `balls__special` is needed since we already have cache, but I'm not 100% sure of that

ballsdex/packages/balls/cog.py
```diff
@@ -660,3 +661,102 @@ async def count(
...
+            source.embed.title = f"Top {len(results)} Duplicate Specials"
+            source.embed.color = discord.Color.purple()
+            source.embed.set_thumbnail(url=interaction.user.display_avatar.url)
+            source.embed.set_footer(
+                text=f"Requested by {interaction.user}",
+                icon_url=interaction.user.display_avatar.url,
+            )
+
+            paginator = Pages(source, interaction=interaction)
+            await paginator.start(ephemeral=True)
```
laggron42, line 762: That's a lot of repeated code from the above block. I understand why you would want separate logic for the queryset part, but for the other code it would be better to have one single block with a variable name changing between special or collectible name.

laggron42 (commented):
I haven't tested yet, this is just a quick review from looking at the new code. Thanks for the contribution, this looks like a useful command!

ballsdex/packages/balls/cog.py
MaxxMXSZ: Yeah I absolutely agree, what's a fair cooldown? a minute?

ballsdex/packages/balls/cog.py
MaxxMXSZ: Tested it. Indeed it is already cached

flaree (approved):

PR #490, opened 2025-01-01
[info] Add /tutorial command

### LINKED ISSUE #377: A command that shows a tutorial for BallsDex!

imtherealF1:
resolves #377


laggron42 (changes requested):

flaree:
I don't believe we need a tutorial command, most commands are self explanatory.

PR #492, opened 2025-01-02
[balls] Handle more DoesNotExist errors

### LINKED ISSUE #486: Player does not exist error on certain commands
**Describe the bug**
Some of the Player.get calls do not handle the DoesNotExist error thrown and thus results in a error being thrown.

**To Reproduce**
1. Use balls list or favourite while not having an account created for the invoking user.

**Expected behavior**
Command to execute as intended

imtherealF1:
resolves #486

### DIFF
--- a/ballsdex/packages/balls/cog.py
+++ b/ballsdex/packages/balls/cog.py
@@ -474,9 +474,7 @@ class Balls(commands.GroupCog, group_name=settings.players_group_cog_name):
 
         if not countryball.favorite:
             try:
-                player = await Player.get(discord_id=interaction.user.id).prefetch_related(
-                    "balls"
-                )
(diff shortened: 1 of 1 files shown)


ballsdex/packages/balls/cog.py
```diff
@@ -158,7 +158,14 @@ async def list(
         if user is not None:
             if await inventory_privacy(self.bot, interaction, player, user_obj) is False:
                 return
-        interaction_player = await Player.get(discord_id=interaction.user.id)
+        try:
+            interaction_player = await Player.get(discord_id=interaction.user.id)
+        except DoesNotExist:
```
flaree, line 163: Instead of saying they should create an account, should we use get_or_create?

flaree (approved):

PR #496, opened 2025-01-03
[utils] Prevent is_staff from erroring if roles are None

### LINKED ISSUE #485: is_staff check can error when roles are None
**Describe the bug**
Most likely a gateway issue however Member.roles seems like it can contain a None type - just needs some specific handling

**To Reproduce**

**Expected behavior**
Commands with is_staff to work as intended

Cannot repo outside of Ballsdex.

imtherealF1:
resolves #485


ballsdex/core/utils/utils.py
```diff
@@ -11,9 +11,10 @@
 
 def is_staff(interaction: discord.Interaction) -> bool:
     if interaction.guild and interaction.guild.id in settings.admin_guild_ids:
-        roles = settings.admin_role_ids + settings.root_role_ids
-        if any(role.id in roles for role in interaction.user.roles):  # type: ignore
-            return True
+        if isinstance(interaction.user, discord.Member):
```
laggron42, line 14: This check isn't enough, from my understanding the issue is that roles returned by the API are `None`, if the base type of the user was `discord.User` to begin with, then the attribute `roles` wouldn't exist at all. I am pretty confident that all users we are dealing with are always `discord.Member` objects, so instead you need to filter out the items from `member.roles` that are empty

imtherealF1:
closing, as it was a ballsdex issue only, and this wont fix it

PR #498, opened 2025-01-03
[ballsdex] Add long comments to most functions

imtherealF1:
makes the code much more understandable and accessible for new coders who may be interested in understanding the functionality of the bot.


flaree:
Personally not a fan of these, adds them to places where its just boilerplate code and doesn't need them - a couple are also wrong

imtherealF1:
may as well close this then, quite a pointless pr now that i thought about it again

PR #501, opened 2025-01-04
[admin] improve /blacklist history, seperate id arg into user + guild_id

imtherealF1:
makes it look cleaner, especially because at the start the argument was named user_id, which didn't fit guilds at all


flaree:
We're going to leave it as is - the only thing that can be changed is the variable name - Right now if the guild no longer exists it also wont show any history. I prefer it as the one approach.

Thanks for the PR

Issue #503, opened 2025-01-06
[trade] Items can be shown as duplicated


laggron42:
**Describe the bug**

When performing a trade, it is possible to duplicate an item in the listing. The same countryball will be listed twice, but the final trade doesn't duplicate the countryball itself. It is a visual but dangerous glitch.

The duplicated item also appears in history.

![image](https://github.com/user-attachments/assets/6909e9ae-a68c-484a-b1be-5778fd160c2b)
![image](https://github.com/user-attachments/assets/60b3271b-1829-4f69-b48b-6db4060f80b7)

**To Reproduce**

> You just click send twice quickly

PR #505, opened 2025-01-07
[player] add another check for removing chance of duped friends

imtherealF1:
this is the same reason as #503

### DIFF
--- a/ballsdex/packages/players/cog.py
+++ b/ballsdex/packages/players/cog.py
@@ -283,9 +283,15 @@ class Player(commands.GroupCog):
         if not view.value:
             self.active_friend_requests[(player1.discord_id, player2.discord_id)] = False
             return
+        else:
+            if friended:
+                await interaction.followup.send(
+                    "You are already friends with this user!", ephemeral=True
+                )
+                return
 
-        await Friendship.create(player1=player1, player2=player2)
-        self.active_friend_requests[(player1.discord_id, player2.discord_id)] = False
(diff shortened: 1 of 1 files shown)


ballsdex/packages/players/cog.py
```diff
@@ -283,9 +283,15 @@ async def friend_add(self, interaction: discord.Interaction, user: discord.User)
...
+            if friended:
+                await interaction.followup.send(
+                    "You are already friends with this user!", ephemeral=True
+                )
+                return
 
-        await Friendship.create(player1=player1, player2=player2)
-        self.active_friend_requests[(player1.discord_id, player2.discord_id)] = False
+            await Friendship.create(player1=player1, player2=player2)
+            self.active_friend_requests[(player1.discord_id, player2.discord_id)] = False
```
flaree, line 294: is this indentation needed?

ballsdex/packages/players/cog.py
```diff
@@ -283,6 +283,12 @@ async def friend_add(self, interaction: discord.Interaction, user: discord.User)
         if not view.value:
             self.active_friend_requests[(player1.discord_id, player2.discord_id)] = False
             return
+        else:
+            if friended:
```
laggron42, line 287: This variable is not refreshed, you're re-using the old result and it can never be true, type checkers even raise a warning.

Have you found a situation where a duplicate is possible here? There seems to be a lot of checks already to avoid sending a friend request twice.

PR #508, opened 2025-01-11
Changes to determine credit color

### LINKED ISSUE #507: Change colour of credits to make it readable depending on the card bg
Analyse the colour of the background and get a colour that will match it to make the credits readable

using the like of scipy or another viable lib

MaxxMXSZ:
resolves #507
Choosing between black or white depending on the color dominant "brightness" around the credit text area.


MaxxMXSZ:
I am not certain if this is what you actually wanted but personally it looks good.
Black and White are basically Neutral color it just go with everything instead of colorful color.

flaree:
please post screenshots showing it in action

MaxxMXSZ:
example 3
example 4
example 2
example1
Here are the examples

laggron42:
I haven't tested in details yet but I'm slightly worried about performances. Do you think you could add a cache on this? Just a simple dictionary with background path as key will be more than enough and save a LOT of CPU time

MaxxMXSZ:
> I haven't tested in details yet but I'm slightly worried about performances. Do you think you could add a cache on this? Just a simple dictionary with background path as key will be more than enough and save a LOT of CPU time

I COULD add a cache yes I completely understand your worries about performance but this process take super fast I would say not even a millisecond I think it is just easier to keep it like this. But in case of keeping cache of the credit color would it also be interesting to add an attribute to change color (Hex) in panel? I am happy to do both but I think keeping it as it is, isn't a problem at all.

flaree:
I think just a cache is fine

MaxxMXSZ:
Have not been tested yet. Would appreciate someone testing it thanks.

flaree (approved):

PR #520, opened 2025-01-25
Add app_commands.MissingRole handler.

russia1848:


ballsdex/core/bot.py
```diff
@@ -414,14 +414,6 @@ async def on_command_error(
                 )
                 return
 
-            if isinstance(exception, commands.MissingPermissions):
```
flaree, line 417: Any reason as to why this was removed, we have quite a few commands with just permission checks

flaree (changes requested):

ballsdex/core/bot.py
russia1848: I mean in admin commands.

russia1848 (commented):
Fix a little typo.

ballsdex/core/bot.py
flaree: Your completely removing the permission handling, which effects every command

ballsdex/core/utils/utils.py
```diff
@@ -37,7 +37,7 @@ async def inventory_privacy(
     elif privacy_policy == PrivacyPolicy.FRIENDS:
         if not await interacting_player.is_friend(player):
             await interaction.followup.send(
-                "This users inventory can only be viewed from users they have added as friends.",
+                "This user's inventory can only be viewed from users they have added as friends.",
```
flaree, line 40: This was correct before, it's now a typo by changing it

flaree:
Thanks for the PR but this just isn't something that's required as of now. We appreciate the effort put in

PR #523, opened 2025-01-27
Change "startup_time" to "uptime" and use Danny's time.human_timedelta.

russia1848:


flaree:
Sorry, but just copy and pasting someone else's code without even attribution here isn't what we're looking for.

PR #524, opened 2025-01-27
DM user at /admin balls give

Nipatsuplayer:
Bot also sends DM to user so user knows that he got new ball with admin command


russia1848:
Good idea.

Nipatsuplayer:
> Sorry but I don't feel this is something necessary that needs to be added, in fact very rarely would I want to dm when using that command.
> 
> Please create a feature request issue beforehand to triage the issue before creating a PR just to gauge interest

Okay! I have this feature in my dex and I need it all the time so decided to do pull request if you would like to add it. But np

PR #531, opened 2025-01-31
Update pyproject to use Python 3.13 because of admin panel

uint128-t:
Python 3.13 is required but pyproject.toml only requires python 3.12


laggron42 (dismissed):

laggron42:
Can you fix CI by running `poetry lock` ? Otherwise I'll do it in a few hours

uint128-t:
my bad but what is CI? poetry lock doesn't seem to modify anything

uint128-t:
ok i did it how do i update the pr?

laggron42 (approved):

PR #533, opened 2025-02-05
[models] change capacity in ball column name to ability

imtherealF1:
Renaming it, avoids confusion especially with new self hosters.

Tasks:
- [x] Rename in every file
- [x] Create migration for django


flaree:
Just a note, all migrations are currently on hold until we get rid of tortoise

PR #535, opened 2025-02-07
Use app_commands.Range in every command and change "config disable" name to "config toggle".

russia1848:


flaree:
Hi, thanks for the PR



However i'm going to deny this based on one part of this PR already existing as an open PR. In future we do ask that you split pull requests up into their separate use cases so that they can be reviewed appropriately


Thank you

PR #541, opened 2025-02-10
[transformers] coalesce both catch names and translations

flaree:


flaree:
confirmed working by f1

PR #543, opened 2025-02-10
Remove "Provide either user or user_id." and change "user_id" to "id".

finland1848:


flaree:
This PR makes no sense bae changing one arg to something more confusing?

PR #545, opened 2025-02-11
[migrations] Fix incorrect file path in migrations

Dotsian:
Regimes, shinies, and economies will still use the old file path. When you clone Ballsdex and try to spawn something, it will output an error.


flaree:
Editing old migrations isn't the best solution, especially one that's very old as they'll never be. ran by anyone who's installed the bot or updated it. The best way is to create a new migration but right now we have a development hold on anything that involves a migration while we fully switch away from Django

Dotsian:
Alright, I'll close this for now.

PR #552, opened 2025-02-19
[Trade] Add trade cooldown bypass policy

flaree:

- Add cooldown bypass policy
- Handle cooldown bypassing in trade
- Add policy to info command (admin/player)

Yes


TridentInsurgence (commented):
Instead of "Bypass acceptance cooldown" maybe call it "No acceptance cooldown" or "Remove acceptance cooldown"

flaree:
I prefer bypass as that's what it is, you're bypassing it. It's on by default.

PR #554, opened 2025-02-19
--use-aerich command.

finland1848:
People might hate that you moved to Django, so let them use Aerich.


laggron42:
This cannot work for a long number of reasons
- Even if you don't use Django for performing the actual migration, its internal state still needs to be updated, or the admin panel won't work. Only updating the state means setting `managed` to `False` in the Django models, but it generates different migrations. It's literally impossible to have two different lines of migrations for those who want to use them or not (unless you have separate apps, it's getting very complex)
- Aerich is literally broken. All migrations have been written by hand for a while because the commands generates you 15 lines of broken index bullshit. It's straight up not working.
- Suppose the two lines above were magically fixed. Even then you're asking us to support two different migration engines, test every case, support it when it obviously breaks, that is *so much* additional work
- This would also mean being limited to Aerich's capabilities (very limited), if we want to use some of the newer migration features allowed by Django, well we can't. Neither can we use Python based migrations. It would just pull us back, and the entire goal of this was to stop being slowed down by this shit
- You're asking way too much from me. I am being nice by publishing the code open source, and taking selfhosters into account when coding new stuff, but my priority is still my own bot. Taking these cases into consideration already slows my development so much. If I didn't have to consider others, this would have been done in a day.
- You just added back the old migration files with some old code pasted, without even being aware of all the fundamental limitations I cited above (the first point is almost impossible to solve)
- Your code doesn't even compile

The only feasible solution is to fork the bot, and support it yourself.

finland1848:
Fair.

PR #556, opened 2025-02-20
Fix permissions checks in config channel.

finland1848:


ballsdex/packages/config/cog.py
```diff
@@ -56,15 +56,34 @@ async def channel(
...
+       
         if channel is None:
             if isinstance(interaction.channel, discord.TextChannel):
                 channel = interaction.channel
             else:
                 await interaction.response.send_message(
                     "The current channel is not a valid text channel.", ephemeral=True
                 )
-                return
+        else:
```
flaree, line 67: Instead of if hell, you could move this into just one check for multiple

flaree (changes requested):

ballsdex/packages/config/cog.py
finland1848: Hmm, okay.

PR #558, opened 2025-02-23
Fix a little mistake in player policy donation's description.

finland1848:


Dotsian:
Oh for Christ's sake

finland1848:
What?

PR #559, opened 2025-02-24
[trade] Add trade_id arg in /trade view

imtherealF1:
Adds a trade_id arg in /trade view, so that a user can check a previous trade they've done.

Yes, except the "if interaction.user.id not in (trade.player1.discord_id, trade.player2.discord_id):" check.

### DIFF
--- a/ballsdex/packages/trade/cog.py
+++ b/ballsdex/packages/trade/cog.py
@@ -8,6 +8,7 @@ from discord import app_commands
 from discord.ext import commands
 from discord.utils import MISSING
 from tortoise.expressions import Q
+from tortoise.exceptions import DoesNotExist
 
 from ballsdex.core.models import BallInstance, Player
 from ballsdex.core.models import Trade as TradeModel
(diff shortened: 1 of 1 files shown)


ballsdex/packages/trade/cog.py
```diff
@@ -427,16 +428,43 @@ async def history(
...
         """
-        trade, trader = self.get_trade(interaction)
-        if not trade or not trader:
-            await interaction.response.send_message(
-                "You do not have an ongoing trade.", ephemeral=True
-            )
-            return
+        if trade_id:
+            try:
+                trade = await TradeModel.get(id=trade_id).prefetch_related("player1", "player2")
```
laggron42, line 443: You forgot to convert from base 16
```suggestion
                trade = await TradeModel.get(id=int(trade_id, 16)).prefetch_related("player1", "player2")
```

ballsdex/packages/trade/cog.py
```diff
@@ -427,16 +428,43 @@ async def history(
...
+            if interaction.user.id not in (trade.player1.discord_id, trade.player2.discord_id):
+                await interaction.response.send_message(
+                    "You were not part of this trade.", ephemeral=True
+                )
+                return
+
+            trader1 = await TradingUser.from_trade_model(trade, trade.player1, self.bot)
+            trader2 = await TradingUser.from_trade_model(trade, trade.player2, self.bot)
+
+            source = TradeViewMenu(interaction, [trader1, trader2], self)
```
laggron42, line 459: This class needs to be modified too. It calls `get_trade` which only works for active trades. When testing, I only got this since the trade wasn't active
image

ballsdex/packages/trade/cog.py
```diff
@@ -427,16 +428,43 @@ async def history(
...
+        if trade_id:
+            try:
+                trade = await TradeModel.get(id=trade_id).prefetch_related("player1", "player2")
+            except DoesNotExist:
+                await interaction.response.send_message(
+                    f"Trade with ID #{trade_id} not found.", ephemeral=True
+                )
+                return
+
+            if interaction.user.id not in (trade.player1.discord_id, trade.player2.discord_id):
```
laggron42, line 450: I was going to say, a call to `is_staff` may be useful, but we already have `/admin history trade`. Both commands have kinda the same purpose, how about deleting it too?

laggron42 (changes requested):
Not exactly working yet but almost good. I think you can delete `/admin history trade` and replace it with this command.

imtherealF1:
> Not exactly working yet but almost good. I think you can delete `/admin history trade` and replace it with this command.

seperate pr or this one?

Death-laso-push:
Fuck off

PR #560, opened 2025-02-24
[translations] Add plural collectible and player group cog translations

Dotsian:

The player cog references the `/balls give` command in the donation function's docstring, which wouldn't be translated if the player group cog name was changed. Also, `countryballs` wouldn't use the correct plural version once translated.

No


finland1848:
#558

PR #576, opened 2025-03-06
Fix bug in balls create.

afrozenlake:

Some people enter an ID of one of their's bots emoji but the bot searches for the emoji in one of the servers that it's in,  therefore the bot says "The bot does not have access to the given emoji.". I fixed that.

Yes.

### DIFF
--- a/ballsdex/packages/admin/balls.py
+++ b/ballsdex/packages/admin/balls.py
@@ -566,10 +566,12 @@ class Balls(app_commands.Group):
             return
         emoji = interaction.client.get_emoji(int(emoji_id))
         if not emoji:
-            await interaction.response.send_message(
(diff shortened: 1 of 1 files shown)


laggron42:
I can see what you're trying to do, but `bot.get_emoji` already accesses the cached application emojis. If the emoji is not found, then you need to run `b.reloadcache`. Since that command reloads the cache anyway, it's not really a bad idea, but you need to follow the coding style plz. Every PR you make has super weird indentation, like it was coded from the phone. Look at why CI is failing, and fix it (4 spaces indentation).

And please, listen to what flare said, you send a LOT of PRs and I really appreciate the effort, but it's also hard on us to review

switzerland1848:
Okay, sorry.

PR #581, opened 2025-03-25
Move capacity description based on length of title

dormieriancitizen:

This is a pretty small change that moves the ability description up one line in the event that the ability name is one line long. 

Example below; details of card hidden because it contains PII and whatnot. The only relevant part is the spacing between the ability name and description, which on the previous fork would be padded an extra 100px.

image

Yes, tested on my fork (along with other changes) and with dummy data on this branch


laggron42 (approved):
Works for me, thanks a lot!

PR #583, opened 2025-03-28
Create slash-drop.py

TheOther1211:
This is basically just another /command-->
/drop lets you drop a ball so it appears as a country ball spawning but will have a custom message mentioning who dropped it and others can catch it as if it were a spawned ball
I never tested this....


TheOther1211:
just check out the file to see lol

TheOther1211:
this is the code btw https://github.com/TheOther1211/BallsDex-DiscordBot/blob/patch-2/slash-drop.py

flaree (changes requested):
Hi -- thank you for the PR.


As it stands this is not in a state that we can use. Its just a file in the root of the repository, you need to add it to the respective cog.

TheOther1211:
> Hi -- thank you for the PR.
> 
> As it stands this is not in a state that we can use. Its just a file in the root of the repository, you need to add it to the respective cog.

I just started using pull requests and I wasn't sure on how to do that sorry. I'll also update the file to work with ballsdex I'm more familiar with discord.js than python discord.

flaree:
You've arguably made it worse than the first attempt. As I mentioned before, you need to place this within one of the existing cogs. You've just left a random file in the root of the repository, which well - does nothing

laggron42:
The thing is, you don't seem to know what you're doing at all. A lot of functions are made-up (like `drop_manager` or `bot.logger`, which is very common with AI code btw), you don't get the difference between `Ball` and `BallInstance` (you'd be deleting every single instance of a country by using the wrong object) and you're not using the tools available in the code base.

Another big problem in the code is how you're selecting the countryball to drop, not only are you using the wrong object type but you're also accepting a raw ID argument (not converted from hexadecimal), and if not given, random ball is chosen, which is a bad idea for multiple reasons. The code base is filled with utilities to make coding end functions almost trivial (for example `BallInstanceTransform` enables auto-completion for selecting a ball, `ConfirmChoiceView` is an easy way to prompt for confirmation...) which you should be using. The one big thing missing is the ability to spawn an existing countryball, but I wanted to refactor that function for a while, I can help with that tbh.

---

Lot of talk, but **please don't try again without changing your methods**, I would love to help you get in the right track. Join the [developer server](https://discord.gg/PKKhee4fvy) and ask me for help in the coding channel, you need to configure the bot locally such that you can test your feature and submit code that works.

PR #586, opened 2025-03-31
Change setup of Docker

dormieriancitizen:

This PR adjusts the setup of the dockerfile & docker compose to do the following

#### Dockerfile changes
- Rebase on Alpine instead of on Debian—it's lighter and builds faster
- Disable pycache; it's not needed in docker and clutters the directories
- Move the initial poetry install to before the code is copied. Because Docker caches builds, this ensures that the build will only re-install dependencies if the poetry file changes; this should make the vast majority of builds significantly faster since we no longer install a heap of dependencies
- Remove poetry cache after build so it isn't included in the final image.
- Remove the 2s delay for postgres (see docker-compose.yml changes)

#### Dockerignore changes
 - This PR adds a dockerignore containing files that shouldn't be embedded in the image
 - For example, before this, any change to the logfile would force a rebuild (which would also re-install dependencies prior to this), which is needlessly wasteful and also bloats the size of the image.
 - Now, only code changes should get docker to rebuild instead of using the cached image.
 
#### Docker-compose.yml changes
 - This restructures the docker-compose.yml to rely upon healthchecks and depends_on for migration and postgres rather than the old method of running a migration for both services before running the service
 - Postgres is given a healthcheck using pg_isready
 - All other services depend on condition: service_healthy, so they will only start once postgres indicates that it is ready to recieve requests.
 - Added a migration service: this runs once, before the admin_panel and ballsdex services, since they depend on service_completed_successfully.
 - Because of this, we can stop running the migration twice (once for admin panel and once for ballsdex) while still ensuring migration takes place before the bot launches.
 - Also, swapped the volume mounts from explicitly saying bind mount to just using ./:/code, since it's the default anyways.

Yes, changes tested on this fork from a clean DB & config (I did not have a used DB to test with, but that should probably be tested before this is merged), and my fork has been running with similar changes for months with no problems.

- [ ] Test on instance that contains data
- [ ] Ensure there are no shell dependencies that Alpine does not have (I didn't see any but you never know)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,6 +3,6 @@ admin_panel/static
 pgbackups
 config.yml
 .env
-ballsdex.log
+ballsdex.log*
 venv
 __pycache__
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,8 @@
 admin_panel/media
 admin_panel/static
+pgbackups
+config.yml
+.env
+ballsdex.log
 venv
 __pycache__
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+admin_panel/media
+admin_panel/static
+venv
+__pycache__
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1
 
-FROM python:3.13.0-bookworm
+FROM python:3.13.0-alpine
 
 ENV PYTHONFAULTHANDLER=1 \
   PYTHONUNBUFFERED=1 \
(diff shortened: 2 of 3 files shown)


dormieriancitizen:
I'd also note that (while this PR does not do that), it might be a good idea to move docker-compose.yml to, ie, docker-compose.yml.example and gitignore docker-compose.yml

That way people can make changes to the docker-compose without causing merge conflicts (ie to change port, add prod django file, etc)

laggron42 (commented):
Wow thank you so much, that's really a great addition, didn't know you could do that migration thing with compose! Just a few nitpicks

Issue #591, opened 2025-04-08
Adjust special algorithm to be more transparent


dormieriancitizen:
The current special algorithm can be a bit confusing, since the way it works effectively halves the chance of any one special spawning when another is added. This is made in the goal of ensuring that if the chances sum to a number greater than one, commons still spawn. 

While this is a noble goal, it causes other problems in the form of having to adjust all the other chances when an event is active, and it being unclear without some math what the actual chances of a given special spawning are (not ideal)

My proposal is instead an algorithm where what you see is what you get, so long as the sum of weights remains under one; ie you put in 0.025 as the weight for a special and it has a 2.5% chance of spawning. This can be done in a pretty simple way:

- Take all active weights and sum them
- Subtract that value from one; if it's negative, set it to 0
- Add that value to weights with special as none
- random.choices with those weights

Also, might be good to pull out get_random_special in a similar way to get_random_countryball

While this will cause all balls to be special if the weights sum above one (and to collapse the weights), this makes sense to me and I think to most users of the admin panel: if the chance of each special spawning sums to above one, commons won't spawn. And it makes it much easier to get the desired spawn chance out of any given special.

Happy to send in a PR later if this would be a welcome change.

dormieriancitizen:
Btw this is a one-line change (summing differences vs difference of sums)

laggron42:
Agreed, I don't like the way it's done right now, the way you described seems really good you can go ahead with that!

`get_random_special` would also be a good addition, perhaps as a static method of `Special` or `BallSpawnView`

dormieriancitizen:
Closed now (it got merged)

PR #596, opened 2025-04-16
Change special algo

dormieriancitizen:

Addresses #591

Basically instead of doing `sum(1-rarity)` to get the weight for commons, we instead do `1-sum(rarities)`.

Idk how to commit to the wiki, but basically the way you convert from the old rarities to the new ones:

- Calculate the old actual spawn change
- This is (rarity value) divided by (number of active specials)
  - note: if you look at the old algorithm as a list of variables, actually when it comes to the division the rarities cancel out but the ones remain, so this is accurate regardless of the set rarities of the specials
 -  Set that calculated chance as the new rarity value

With this change if you want a special to have say a 20% change of spawning you just set the rarity to 0.2

Tested locally, and my fork has been running a functionally identical change for a while. Before merging into a big dex like ballsdex with a lot of specials you should probably make sure that the odds are as you want first, though, I don't have enough specials to test thoroughly.


dormieriancitizen:
Also unrelated side tangent but upon looking at the pre-commit check ballsdex runs black twice
image

flaree (approved):

PR #600, opened 2025-04-18
[admin panel] Use configured bot name instead of "BallsDex"

imtherealF1:

Use settings.bot_name instead of "BallsDex"

Yes

### DIFF
--- a/admin_panel/admin_panel/admin.py
+++ b/admin_panel/admin_panel/admin.py
@@ -1,8 +1,10 @@
 from django.contrib import admin
 
+from ballsdex.settings import settings
+
 
 class BallsdexAdminSite(admin.AdminSite):
-    site_header = "Ballsdex administration"  # TODO: use configured bot name
-    site_title = "Ballsdex admin panel"
+    site_header = f"{settings.bot_name.title()} administration"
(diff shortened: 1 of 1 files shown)


admin_panel/admin_panel/admin.py
```diff
@@ -1,8 +1,10 @@
 from django.contrib import admin
 
+from ballsdex.settings import settings
+
 
 class BallsdexAdminSite(admin.AdminSite):
-    site_header = "Ballsdex administration"  # TODO: use configured bot name
-    site_title = "Ballsdex admin panel"
+    site_header = f"{settings.bot_name.title()} administration"
+    site_title = f"{settings.bot_name.title()} admin panel"
```
flaree, line 8: I don't think we need to use title here as that can be done directly via the config setting.

```suggestion
    site_header = f"{settings.bot_name} administration"
    site_title = f"{settings.bot_name} admin panel"
```

flaree (changes requested):

laggron42 (approved):
Interesting, I didn't think `settings` variable was available by the time this class was loaded (which is why I didn't even try). Tested and works, just apply flare's suggestion and we're good to merge

flaree (approved):

PR #601, opened 2025-04-23
Remove tortoise

laggron42:
## Description of the changes

Completely remove the Tortoise ORM, replaced with Django.

### Why?

- Tortoise is not truly asynchronous and blocks the main loop too often, severely blocking performances, see https://github.com/tortoise/tortoise-orm/issues/1378
- Some advanced features are lacking. They make some progress (https://github.com/tortoise/tortoise-orm/issues/1612 got a PR :D) but there's still a lot missing compared to Django
- The ecosystem around it is too weak, we already moved to Django migrations and admin panel for that reason
- There are two separate `models.py` files to maintain because of that, it's time to make that one

Big thanks to the Tortoise team for their work, I still think it's a good library for small to medium scale projects, we just grew out of this scope.

### What's changing in the code

- All imports from `ballsdex.core.models` have to be replaced with `bd_models.models`. The class names are the same.
- When fetching objects, the manager is accessed with `Model.objects`. Methods also have to be prefixed with `a` (for async)[^1]
  | Before                                       | After                                                      |
  |----------------------------------------------|------------------------------------------------------------|
  | `await Ball.get(id=1)`                       | `await Ball.objects.aget(id=1)`                            |
  | `await BallInstance.exists(player_id=1)`     | `await BallInstance.objects.filter(player_id=1).aexists()` |
  | `await Player.get_or_create(discord_id=...)` | `await Player.objects.aget_or_create(discord_id=...)`      |
- Unlike Tortoise that made it clear when the query was executed (when it's awaited), Django does lazy querying, meaning the query only runs when needed. There is no `afilter()` that runs the query and compiles it in a list, the query has to be iterated to run.
  - The recommended way to iterate over queries is using `async for` loops[^2]
  - A method `aall()` has been added as a workaround to compile a query into a list, but it should be avoided if possible

### New risks

One big risk of using the Django ORM is that it supports both sync and async, which can create odd situations with both queries mixed. It's very important that all queries done in the bot are async, or it will cut off the connection to Discord.

The biggest risk is related objects. On Tortoise, you are forced to add `prefetch_related` or `select_related` to explicitly confirm you need additional objects retrieved. With Django, you can omit those, and each time a related object is requested, it creates a new syncrhonous query[^3] (if not prefetched), which is really bad and hardly detectable!

```py
>>> # using Tortoise
>>> b = await BallInstance.get(id=0x42)
>>> b.ball.country
AttributeError: 'QuerySet' object has no attribute 'country'.

>>> # using Django
>>> b = await BallInstance.objects.aget(id=0x42)
>>> b.ball.country
"France"
>>> b.ball.health
20
>>> b.ball.attack
20
>>> # 4 queries (3 synchronous) have just been made here !!
```

It's very hard to detect, I'll look into profiling tools to check the number of queries performed.

### Additional notes

- Django does not provide `get_or_none`. To avoid rewriting too much code, it has been reimplemented in the default manager
- Some additional managers have been added for convenience, such as `Ball.enabled` for returning only enabled balls
- The whole bot needs to be tested. Every single command and feature is being modified in this PR
- The bot's entrypoint moved to be a Django management command. To start the bot, you would have to do `python3 manage.py startbot`, but because of the weird file architecture, `python3 -m ballsdex` will still work and correctly edit working directories.
  - Some working directories will be incorrectly returned, such as `ball.wild_card.path`

[^1]: https://docs.djangoproject.com/en/5.1/topics/db/queries/#async-queries
[^2]: https://docs.djangoproject.com/en/5.1/topics/db/queries/#query-iteration
[^3]: https://docs.djangoproject.com/en/5.1/ref/models/querysets/#prefetch-related


laggron42:
Merging now even tho it's probably still bugged. All 3.0 dev will be based off this so it's better to merge early, and bug fix later (when 3.0 branch will be ready as a whole)

PR #608, opened 2025-05-03
Configurable spawn messages

dormieriancitizen:

This PR allows users to configure spawn, catch, slow, and wrong messages within the config.yml

I didn't make it configurable from the DB because a) I don't think people are changing this that much b) db calls are slow (comparatively) and there are a lot of messages sent and lastly and most importantly c) I don't want to deal with DB nonsense

Users change format strings in the config.yml, and the bot picks a random one, formats it, and sends it instead of the prior hardcoded strings. It provides the ball name, the collectible singular and plural, and a mention for the user. I did not weight the messages, because I don't think I would use that and can't really think of a usecase.

I think I did all the yml things right, but I'm not sure, because that thing is a terrifying agglomeration of every change to the config since Ballsdex began.

This shouldn't change the user experience for ballsdex users at all (unless you want to) but makes it easier for other dexes to configure and differentiate their dex from stock bd.

Yes, tested for each message type, with 2 messages (responded randomly and seemingly correctly). My fork has had a similar feature for several months, and I've not had problems there either.

### DIFF
--- a/ballsdex/packages/countryballs/countryball.py
+++ b/ballsdex/packages/countryballs/countryball.py
@@ -89,7 +89,6 @@ class CountryballNamePrompt(Modal, title=f"Catch this {settings.collectible_name
             )
             return
 
-        # Good to catch ball now
         ball, has_caught_before = await self.view.catch_ball(
             interaction.user, player=player, guild=interaction.guild
         )
--- a/ballsdex/packages/countryballs/countryball.py
+++ b/ballsdex/packages/countryballs/countryball.py
@@ -60,29 +60,45 @@ class CountryballNamePrompt(Modal, title=f"Catch this {settings.collectible_name
 
         player, _ = await Player.get_or_create(discord_id=interaction.user.id)
         if self.view.caught:
+            slow_message = random.choice(settings.slow_messages).format(
+                user=interaction.user.mention,
+                collectible=settings.collectible_name,
+                ball=self.view.name,
+                collectibles=settings.plural_collectible_name,
+            )
+
             await interaction.followup.send(
-                f"{interaction.user.mention} I was caught already!",
+                slow_message,
                 ephemeral=True,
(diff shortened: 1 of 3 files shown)


ballsdex/packages/countryballs/countryball.py
```diff
@@ -60,29 +60,45 @@ async def on_submit(self, interaction: discord.Interaction["BallsDexBot"]):
...
-        else:
-            await interaction.followup.send(
-                f"{interaction.user.mention} Wrong name!",
+                wrong_message,
                 allowed_mentions=discord.AllowedMentions(users=player.can_be_mentioned),
                 ephemeral=False,
             )
+            return
+
+        # Good to catch ball now
```
flaree, line 92: Just a pet peeve, comments with no use should be removed

flaree (changes requested):

flaree (changes requested):
I ran into a general issue, on the first migration all the values will be an empty list. I've replicated this a few times, so I believe extra handling is needed somewhere. After subsequent restarts, it's fine.

Screenshot 2025-05-04 at 09 45 28

dormieriancitizen:
Hm, I'll look into it

flaree (approved):
Thanks for the PR!

Very much appreciated

dormieriancitizen:
Ok, that should fix it. Before, config changes were always applied after reading the config file. I moved the update config line so that it happens before, so that consumers of the settings can always trust that it contains fields they need before using them, that way they don't have to check themselves.

I don't know if there's a reason for this order of read -> update, but I hope I haven't clobbered something important

dormieriancitizen:
Is this something I should be concerned about? I can't see the error since I'm not on BD's sentry.

flaree:
> Is this something I should be concerned about? I can't see the error since I'm not on BD's sentry.

It's fine, it's trying to make a guess on what caused the error but it's just one we get thousands of times a day

PR #611, opened 2025-05-08
Let the wrong name message say what the submitter did wrong

dormieriancitizen:

This lets the wrong_message field contain a {wrong} replacement, which gets filled with whatever the user inputted incorrectly in the catch view.

Also adds some documentation in the config yaml regarding what each field does.

Yep

### DIFF
--- a/ballsdex/packages/countryballs/countryball.py
+++ b/ballsdex/packages/countryballs/countryball.py
@@ -75,12 +75,19 @@ class CountryballNamePrompt(Modal, title=f"Catch this {settings.collectible_name
             return
 
         if not self.view.is_name_valid(self.name.value):
+            if len(self.name.value) > 500:
+                # seems like about the right number
+                wrong_name = "Sorry, entered name too long"
+            else:
+                wrong_name = self.name.value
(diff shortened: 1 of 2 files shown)


ballsdex/packages/countryballs/countryball.py
```diff
@@ -75,11 +75,17 @@ async def on_submit(self, interaction: discord.Interaction["BallsDexBot"]):
             return
 
         if not self.view.is_name_valid(self.name.value):
+            if len(self.name.value) > 500:
+                wrong_name = "Sorry, entered name too long"
```
flaree, line 79:
```suggestion
                wrong_name = self.name.value[:500] + "..."
```

flaree (changes requested):
Just a thought, normally in these instances you just slice the value and return it with an elipsis to signify its omitting text. what do you think?

dormieriancitizen:
Oh, yeah, that seems like a better way to do it (though hopefully no one's putting 500 characters into the textbox)

flaree (approved):

PR #614, opened 2025-05-09
add ephemeral to collection

HaapsaluYT:

Added an optional choice to hide the collection command, if you don't want to show it to others - I personally would find this useful for checking what balls I have (instead of running count several times)

Yes. It has not been tested on a large database but the change is minor so that probably shouldn't matter

### DIFF
--- a/ballsdex/packages/balls/cog.py
+++ b/ballsdex/packages/balls/cog.py
@@ -885,12 +885,12 @@ class Balls(commands.GroupCog, group_name=settings.players_group_cog_name):
         countryball: Ball
             The countryball you want to see the collection of
         hide: bool
-            Whether or not to hide the command from others
+            Whether or not to hide the command from other others
         """
         eph = False
         if hide:
             eph = True
-        await interaction.response.defer(thinking=True,ephemeral=eph)
+        await interaction.response.defer(thinking=True, ephemeral=eph)
         player, _ = await Player.get_or_create(discord_id=interaction.user.id)
 
         query = BallInstance.filter(player=player).prefetch_related(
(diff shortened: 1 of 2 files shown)


ballsdex/packages/balls/cog.py
```diff
@@ -875,6 +875,7 @@ async def collection(
         self,
         interaction: discord.Interaction["BallsDexBot"],
         countryball: BallEnabledTransform | None = None,
+        hide: bool = False,
```
flaree, line 878:
```suggestion
        ephemeral: bool = False,
```

ballsdex/packages/balls/cog.py
```diff
@@ -883,8 +884,13 @@ async def collection(
         ----------
         countryball: Ball
             The countryball you want to see the collection of
+        hide: bool
+            Whether or not to hide the command from other others
```
flaree, line 888:
```suggestion
        ephemeral: bool
            Whether or not to send the command ephemerally.
```

ballsdex/packages/balls/cog.py
```diff
@@ -883,8 +884,13 @@ async def collection(
...
         """
-        await interaction.response.defer(thinking=True)
+        eph = False
+        if hide:
+            eph = True
+        await interaction.response.defer(thinking=True, ephemeral=eph)
```
flaree, line 893:
```suggestion
        await interaction.response.defer(thinking=True, ephemeral=ephemeral)
```

flaree (changes requested):
Hey, thanks very much for the PR.

Just some changes to the car name along with removing the conditional statement as the arg is already defined in the command

flaree (approved):

PR #615, opened 2025-05-09
Validate config file

dormieriancitizen:

This type checks the config.yml to make sure all the things are what they are supposed to be. I'm using typeguard for this, which seems like a fairly robust package, but I'm happy to switch it out for something else if you have suggestions. Also, right now, this exits the program if the check fails, but it could also be made into a warning if that's better.

Also added a fallback for client_id and client_secret since during testing I noticed they were being assigned to none.

I'm not super satisfied with where this is running—I think it would be better to put it in the migration container before the bot or panel runs—but I'm not sure how I would implement that.

Yes


dormieriancitizen:
By the way, the reason I can't use simple isinstance for this is because isinstance doesn't support, for example, `list[int]`; it can only check the general type.

ballsdex/settings.py
```diff
@@ -183,10 +185,10 @@ def read_settings(path: "Path"):
...
 
     if admin := content.get("admin-panel"):
-        settings.webhook_url = admin.get("webhook-url")
-        settings.client_id = admin.get("client-id")
-        settings.client_secret = admin.get("client-secret")
-        settings.admin_url = admin.get("url")
+        settings.webhook_url = admin.get("webhook-url") or ""
+        settings.client_id = admin.get("client-id") or ""
+        settings.client_secret = admin.get("client-secret") or ""
+        settings.admin_url = admin.get("url") or ""
```
flaree, line 191:
```suggestion
        settings.webhook_url = admin.get("webhook-url", "")
        settings.client_id = admin.get("client-id", "")
        settings.client_secret = admin.get("client-secret", "")
        settings.admin_url = admin.get("url", "")
```

Going to leave the full review to laggron, just for this - get has a default value as an arg instead of having to use a conditional

dormieriancitizen:
I'll work on this again once TOML comes out (maybe I'll use json-schema if that's preferable) but for now I'm closing this.

PR #616, opened 2025-05-11
Remove network

dormieriancitizen:

The default behavior of docker compose is to create a single network containing all containers in a compose.yaml; thus there is no need to explicitly create a network. Also, explicitly creating a network makes it more difficult to run multiple dexes, since the internal admin panel / postgres ports collide; this fixes that.

Yes, also tested migrating one that had a network to not having one and no extra stuff was needed.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,8 +20,6 @@ services:
         condition: service_healthy
     # ports:
     #   - "15260:15260"
-    networks:
-      - internal
     volumes:
       - "./:/code"
     tty: true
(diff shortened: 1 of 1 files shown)


laggron42:
Interesting, didn't know docker compose already made its own internal network

> Also, explicitly creating a network makes it more difficult to run multiple dexes, since the internal admin panel / postgres ports collide; this fixes that.

Can you detail that? I run ~80 dexes this way and never had an issue, other than the exposed admin panel port I have to switch, but that's a port collision on the host anyway, no way to get around that

dormieriancitizen:
Actually, now that you mention it, you're right; there wouldn't be a collision because the containers would have different external IPs. 

I hadn't tested that, it was just intuition, which as it turns out, was wrong.

In that case I guess it's just your personal preference as to whether or not to keep the network.

PR #618, opened 2025-05-13
Prevent (some) wrong name abuse

dormieriancitizen:

Prevents people from, for example, pinging @everyone in the wrong name message, when that variable is used

Yes

image


flaree:
I believe these are actually denied by default and don't need to be explicitly declared

dormieriancitizen:
I can confirm they are not denied by default since when I tried it without these changes it pinged everyone

dormieriancitizen:
Example:
image

PR #619, opened 2025-05-14
Hot reload config file

dormieriancitizen:

This adds a helper command to hot reload the config file. Mostly I've been using this to change spawn messages (which hot reload as expected), but I guess it might be useful for other things as well.

I've heard that BD will switch to TOML for 3.0, but as long as the read_settings function still updates the variables in the settings class, this code should still work fine.

Yes


laggron42 (dismissed):
Works for me, just a few nitpicks

flaree (approved):

PR #629, opened 2025-06-16
add /admin return_traded_balls

Ray-Hsueh:

Not sure it's a good idea for a big dex like ballsdex, but this command allows the administrator to take back the ball that the blacklisted person gave away before they died. 
I've tested it on my own dex, and although it seems odd to use days, I have no other ideas at the moment.

Yes

### DIFF
--- a/ballsdex/packages/admin/cog.py
+++ b/ballsdex/packages/admin/cog.py
@@ -1,12 +1,14 @@
 from collections import defaultdict
 from typing import TYPE_CHECKING, cast
+import datetime
 
 import discord
 from discord import app_commands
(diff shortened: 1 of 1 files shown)


flaree:
You need to also go back through every trade with other people and return stuff that was given to other people. 

Realistically, it's an impossible challenge and the reason as to why we haven't added anything related to it.

Ray-Hsueh:
> You need to also go back through every trade with other people and return stuff that was given to other people.
> 
> Realistically, it's an impossible challenge and the reason as to why we haven't added anything related to it.

Didn't really think of that...

PR #630, opened 2025-06-24
Auto-lower command names

dormieriancitizen:

According to my sources
image

So this PR auto-lowers them and removes spaces in order to not break things if the config is strange. Ideally this would be resolved in a config validation stage but this is good in the interim until config gets overhauled with 3.0

Yes


ContestedWheel:
i am the author of the message in the screenshot and can confirm that if you capitalize the following:

- `collectible-name`
- `plural-collectible-name`
- `players-group-cog-name`

you will get a bunch of "Command name is invalid" errors.

Screenshot 2025-06-23 215129

PR #633, opened 2025-07-05
Let users edit the spawn range within config file 

Mitoooooooopo:

this pr simply allows users to config the spawn ranges directly within the config

Yes

### DIFF
--- a/admin_panel/admin_panel/pipeline.py
+++ b/admin_panel/admin_panel/pipeline.py
@@ -168,7 +168,9 @@ async def configure_status(
         # check if user owns the application, or is part of the team and team members are co owners
         async with session.get("applications/@me") as resp:
             info = await resp.json()
-            if info["owner"]["id"] == uid:
+            if info["owner"]["id"] == uid or (
+                info["team"] and info["team"]["owner_user_id"] == uid
+            ):
                 await assign_status(request, response, user, Status.OWNER)
                 return
             if (
--- a/admin_panel/admin_panel/webhook.py
+++ b/admin_panel/admin_panel/webhook.py
@@ -27,8 +27,9 @@ async def notify_admins(
 
     Set `wait` to `False` to ignore the resulting messsage, or failures to send it.
     """
+    log.info(f"Admin notification: {message}")
     if not settings.webhook_url:
-        log.warning(f"Discord webhook URL not configured, attempted to send: {message}")
+        log.warning("Discord webhook URL not configured")
         return
     async with aiohttp.ClientSession() as session:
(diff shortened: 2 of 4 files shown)


ballsdex/settings.py
```diff
@@ -435,6 +444,11 @@ def update_settings(path: "Path"):
   - ballsdex.packages.trade
 """
 
+    if add_spawn_chance_range::
```
laggron42: That doesn't compile
```suggestion
    if add_spawn_chance_range:
```

ballsdex/settings.py
```diff
@@ -435,6 +444,11 @@ def update_settings(path: "Path"):
...
+        content += """
+# range of spawn chance.
```
laggron42, line 449: Same thing here
```suggestion
# spawn chance range
# with the default spawn manager, this is *approximately* the min/max number of minutes
# until spawning a countryball, before processing activity
```

ballsdex/settings.py
```diff
@@ -321,7 +326,10 @@ def write_default_settings(path: "Path"):
 prometheus:
   enabled: false
   host: "0.0.0.0"
-  port: 15260
+  port: 15260 
+  
+# range of spawn chance.
```
laggron42, line 331: Let's add a little more documentation there
```suggestion
# spawn chance range
# with the default spawn manager, this is *approximately* the min/max number of minutes
# until spawning a countryball, before processing activity
```

laggron42 (commented):
Thanks a lot for the contribution! I tested myself and it works (with the compilation error fixed lol)

Could you also add the new entry in `json-config-ref.json` with the following? I can't make suggestions on files not edited yet
```json
        "spawn-chance-range": {
            "type": "array",
            "description": "Spawn chance range. Minimum and maximum values in minutes for the spawn manager to determine when the next countryball should spawn (before taking into account activity)",
            "prefixItems": [
                {
                    "type": "integer",
                    "minimum": 0
                },
                {
                    "type": "integer",
                    "minimum": 0
                }
            ],
            "items": false
        }
```

ballsdex/settings.py
laggron42: Oh well this was fixed already

laggron42:
Also I noticed you changed all line endings, try to configure your editor to have LF and not CRLF, this is what causes the failing CI check. Running black should fix those too

Mitoooooooopo:
oh OK thank you I try to do it right now

ballsdex/settings.py
Mitoooooooopo: I think I done almost

PR #636, opened 2025-07-09
Card templating

dormieriancitizen:

First draft (it works tho) of card templating. Sample output:

image

Yes, but I've not made models yet, so right now the template is a giant dict at the top of image_gen.py. Still better than what was there before I think

I've not checked for performance or error conditions yet, looking for feedback on the system before doing that.


dormieriancitizen:
Ok, this is now at a basic level of usability. Only feature from stock BD that isn't currently replicatable is auto-credits-color—it would be trivial to get it to work but I can't think of a good caching solution atm. But it allows for different templates for specials, regimes, and also configurable fonts and whatnot.

I have not added any way to upload fonts yet but I may do that at some point.

dormieriancitizen:
Not updated for 3.0, will potentially fix later

dormieriancitizen:
Closing in favour of #740

PR #638, opened 2025-07-14
Update libraqm repo

dormieriancitizen:

Builds are currently broken because libraqm is in edge community now, not edge testing. So this PR switches to that repo. 

Yes

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,4 +5,5 @@ config.yml
 .env
 ballsdex.log*
 venv
+.venv
 __pycache__
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,10 +16,10 @@ ENV PYTHONFAULTHANDLER=1 \
 
 # Pillow runtime dependencies
 # TODO: remove testing repository when alpine 3.22 is released (libraqm is only on edge for now)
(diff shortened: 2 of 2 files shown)


dormieriancitizen:
Actually, testing further by starting the image and not just building, something is strange—I'm getting a issue where migration can't import ballsdex? However, I'm also getting this issue without changing the dockerfile, so I suspect something is wrong on my end (maybe the cache is busted).

Regardless, this should be tested on your machine before being merged—while I can't see how this would cause issues like the ones I'm seeing, computers work in mysterious ways.

laggron42:
For some reason you replaced `libraqm-dev` with `libraqm`, the distinction is important because the latter doesn't include the C headers, needed by pillow.

You can test if your dockerfile works by using `python3 -m PIL`, it should show support for all features. Right now, RAQM is missing:
```
$ docker run --rm -it bd-test python3 -m PIL
...
*** RAQM (Bidirectional Text) support not installed
...
```

laggron42 (approved):
I directly edited the file to remove the two `RUN` instructions and combine them as one, allowing lower layer sizes.

PR #647, opened 2025-08-13
configurable catch button

Ray-Hsueh:
Someone said so, so I did it.

Yes

### DIFF
--- a/ballsdex/settings.py
+++ b/ballsdex/settings.py
@@ -130,6 +130,8 @@ class Settings:
     spawn_messages: list[str] = field(default_factory=list)
     slow_messages: list[str] = field(default_factory=list)
 
+    catch_button_label: str = "Catch me!"
+
 
 settings = Settings()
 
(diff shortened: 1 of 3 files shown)


ballsdex/settings.py
```diff
@@ -347,6 +350,9 @@ def write_default_settings(path: "Path"):
   # {user} is mention. {collectible} is collectible name. {ball} is ball name, and 
   # {collectibles} is collectible plural.
 
+  # the label shown on the catch button
+  button_label: "Catch me!"
```
flaree, line 354: Just a nitpick, the comment above may not always be generated if someone is using an old version etc. So the name button_label on its own may be a bit vague, maybe changing it to something more exact like catch_button_label

flaree (changes requested):

flaree (approved):

flaree:
Thanks for the PR!

PR #648, opened 2025-08-15
Fix for auto lower translations bug

AngerRandom:

There's a small bug in "auto lower translations" change, it also applies that to everything (command descriptions, options, etc.). I made a fix for it, but not sure will it work.

No


AngerRandom:
I made a final fix and this time I tested it, it works properly.

flaree (approved):
Thanks for the PR <3

PR #659, opened 2025-09-20
Force wild_card option when creating a countryball via admin commands

HiboMan:

No

### DIFF
--- a/ballsdex/packages/admin/balls.py
+++ b/ballsdex/packages/admin/balls.py
@@ -560,18 +560,18 @@ class Balls(app_commands.Group):
         *,
         name: app_commands.Range[str, None, 48],
         regime: RegimeTransform,
+        economy: EconomyTransform,
         health: int,
         attack: int,
         emoji_id: app_commands.Range[str, 17, 21],
         capacity_name: app_commands.Range[str, None, 64],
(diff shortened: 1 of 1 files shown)


ballsdex/packages/admin/balls.py
```diff
@@ -521,18 +521,18 @@ async def balls_create(
...
+        economy: EconomyTransform,
         health: int,
         attack: int,
         emoji_id: app_commands.Range[str, 17, 21],
         capacity_name: app_commands.Range[str, None, 64],
         capacity_description: app_commands.Range[str, None, 256],
+        wild_card: discord.Attachment,
         collection_card: discord.Attachment,
         image_credits: str,
-        economy: EconomyTransform | None = None,
```
flaree, line 531: Economy should stay as optional, it doesn't break anything. We even have stuff as None in Ballsdex

flaree (changes requested):

ballsdex/packages/admin/balls.py
HiboMan: I understand, thank you

flaree:
I'm actually going to close this PR. We actually use the no spawn image set when creating a ball, so it's there intentionally.

However, thank you for creating the PR

PR #660, opened 2025-09-23
Backport UV to master

dormieriancitizen:

This PR backports UV to master.

Mostly it's because I wanted to figure out how cherry-pick worked, but also it's because poetry builds are crazily slow.

There was a subtle bug I had to fix, which I suspect effects 3.0-dev as well, though I'm not 100% sure
- Because the docker build didn't use the --active flag on uv sync, it placed deps into a .venv in the /code dir
- When it copied over /opt/venv, the image had no dependencies installed (migration then fails to start because it can't import django)
- I've not built 3.0 yet, so it's possible this bug doesn't happen there, but if so I have no idea why it would be different

Yes, build works correctly, bot launches fine. It's way faster also.


dormieriancitizen:
3.0 is out (yay!), will close this

PR #661, opened 2025-09-25
Allow logging with webhooks and remove `log_action`

### LINKED ISSUE #655: Reword log_action to use logging and webhooks
There are currently a lot of duplicated logging actions, to both log it in the console and send a message in the configured channel (which is tricky with clustering btw)

This could be rewritten to use the webhook all the time, remove `log-channel` from config, and use a [custom logging handler](https://docs.python.org/3/library/logging.html#handler-objects) to emit certain messages.

Dotsian:

This pull request refactors action logging to support webhook logging instead of channels. It introduces a `webhook_logger` function, which automatically attaches a `WebhookLoggingHandler` handler and returns the `Logger`.

Additionally, since `log-channel` is deprecated, config migrations will automatically remove the variable from the config file in favor of `webhook-url`. It will also send a warning indicating that channel logging is deprecated if `log-channel` is found.

Resolves #655 

Yes, below is a list of what I have tested.

- [x] Sending a log using `webhook_logger`.
- [x] Checking if `log-channel` is properly removed if it is detected in the config file.
- [x] Attempting to send a log with an invalid webhook.
- [x] Sending a log without a webhook (only sends through console).
- [x] Sending a log that can't be sent through a webhook, such as an empty message.

### DIFF
--- a/ballsdex/core/utils/logging.py
+++ b/ballsdex/core/utils/logging.py
@@ -1,22 +1,73 @@
+import asyncio
 import logging
 
+import aiohttp
 import discord
 
-from ballsdex.core.bot import BallsDexBot
 from ballsdex.settings import settings
 
-log = logging.getLogger("ballsdex.packages.admin.cog")
+log = logging.getLogger("ballsdex.core.utils.logging")
+log.setLevel(logging.INFO)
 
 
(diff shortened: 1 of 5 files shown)


Dotsian:
Whoops, wrong branch, I'll reopen this in a sec

PR #664, opened 2025-10-01
[config] Add admin-channel-ids to config

flaree:

Adds a new entry to admin config, this is for privacy bypassing - meaning staff can only bypass private inventories in a specific channel(s). If left empty, current behaviour is applied

Yes
No

### DIFF
--- a/ballsdex/settings.py
+++ b/ballsdex/settings.py
@@ -99,6 +99,7 @@ class Settings:
     admin_guild_ids: list[int] = field(default_factory=list)
     root_role_ids: list[int] = field(default_factory=list)
     admin_role_ids: list[int] = field(default_factory=list)
+    admin_channel_ids: list[int] = field(default_factory=list)
 
     log_channel: int | None = None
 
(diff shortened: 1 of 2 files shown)


ballsdex/settings.py
```diff
@@ -167,6 +168,7 @@ def read_settings(path: "Path"):
     settings.admin_guild_ids = content["admin-command"]["guild-ids"] or []
     settings.root_role_ids = content["admin-command"]["root-role-ids"] or []
     settings.admin_role_ids = content["admin-command"]["admin-role-ids"] or []
+    settings.admin_channel_ids = content["admin-command"].get("channel-ids", []) or []
```
dormieriancitizen, line 171:
```suggestion
    settings.admin_channel_ids = content["admin-command"].get("admin-channel-ids", []) or []
```

The var, json-schema, and template config file have it as admin-channel-ids, so I presume you meant this (otherwise it doesn't load the channel ids)

json-config-ref.json
```diff
@@ -146,6 +146,16 @@
                     "items": {
                         "$ref": "#/$defs/discord-id"
                     }
+                },
+                "admin-channel-ids": {
+                    "type": [
+                        "array",
+                        "null"
+                    ],
+                    "description": "IDs of channels in which privacy settings can be bypasses. Empty means any channel is allowed.",
```
dormieriancitizen, line 155:
```suggestion
                    "description": "IDs of channels in which privacy settings can be bypassed. Empty means any channel is allowed.",
```

PR #666, opened 2025-10-16
Add button showing recent catches in /admin info user

### LINKED ISSUE #509: [admin] recently caught button on info iser
A recently caught button on the info user menu, when pressed will bring up a menu of the last X catches along with their catch times and server in which they were caught

uint128-t:

Add button in /admin info user to show recent catches of player, along with catch times

Yes

Resolves #509

### DIFF
--- a/ballsdex/packages/admin/info.py
+++ b/ballsdex/packages/admin/info.py
@@ -27,7 +27,7 @@ class PlayerInfoView(discord.ui.View):
         # Display the last 10 catches of the user, and how long it took for each catch
         recent_balls = (
             await BallInstance.filter(
-                player=self.player,
+                player=self.player, spawned_time__isnull=False, trade_player=None
             )
             .order_by("-catch_date")
             .limit(10)


ballsdex/packages/admin/info.py
```diff
@@ -16,6 +16,33 @@
...
+        super().__init__()
+        self.player = player
+        self.username = username
+
+    @discord.ui.button(label="Recent Catches", style=discord.ButtonStyle.primary)
+    async def recently_caught(self, interaction: discord.Interaction, button: discord.ui.Button):
+        # Display the last 10 catches of the user, and how long it took for each catch
+        recent_balls = (
+            await BallInstance.filter(
+                player=self.player,
```
laggron42: Better filter the self-caught only. Also spawned time can be null (admin given), which results in an error, you also need to filter that out.
```suggestion
                player=self.player, spawned_time__isnull=False, trade_player=None
```

uint128-t:
my bad i didn't see this i thought this got

laggron42 (approved):
Thanks!

PR #668, opened 2025-10-20
Allow logging with webhooks and remove `log_action`

### LINKED ISSUE #655: Reword log_action to use logging and webhooks

Caylies:

This pull request refactors action logging to support webhook logging instead of channels. It introduces a `webhook_logger` function, which automatically attaches a `WebhookLoggingHandler` handler and returns the `Logger`.

Additionally, since `log-channel` is deprecated, config migrations will automatically remove the variable from the config file in favor of `webhook-url`. It will also send a warning indicating that channel logging is deprecated if `log-channel` is found.

Resolves #655 

Yes, below is a list of what I have tested.

- [x] Sending a log using `webhook_logger`.
- [x] Checking if `log-channel` is properly removed if it is detected in the config file.
- [x] Attempting to send a log with an invalid webhook.
- [x] Sending a log without a webhook (only sends through console).
- [x] Sending a log that can't be sent through a webhook, such as an empty message.

### DIFF
--- a/ballsdex/settings.py
+++ b/ballsdex/settings.py
@@ -102,8 +102,6 @@ class Settings:
     admin_role_ids: list[int] = field(default_factory=list)
     admin_channel_ids: list[int] = field(default_factory=list)
 
-    log_channel: int | None = None
-
     team_owners: bool = False
     co_owners: list[int] = field(default_factory=list)
 
@@ -169,8 +167,6 @@ def read_settings(path: "Path"):
     settings.root_role_ids = content["admin-command"]["root-role-ids"] or []
     settings.admin_role_ids = content["admin-command"]["admin-role-ids"] or []
     settings.admin_channel_ids = content["admin-command"].get("admin-channel-ids", []) or []
-
-    settings.log_channel = content.get("log-channel", None)
 
     settings.prometheus_enabled = content["prometheus"]["enabled"]
     settings.prometheus_host = content["prometheus"]["host"]
@@ -302,9 +298,6 @@ admin-command:
   # list of channel IDs where admins can bypass privacy settings, empty means no restriction
   admin-channel-ids:
 
-# log channel for moderation actions
-log-channel:
-
 # manage bot ownership
 owners:
   # if enabled and the application is under a team, all team members will be considered as owners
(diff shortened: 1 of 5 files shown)


ballsdex/settings.py
```diff
@@ -549,6 +546,19 @@ def update_settings(path: "Path"):
...
+            "Channel logging is no longer supported. "
+            "Please ensure your 'webhook-url' setting is valid to continue using Discord logging."
+        )
+
+        content = "\n".join(
+            line
+            for line in content.splitlines()
+            if not line.strip().startswith("log-channel:")
+            and not line.strip().startswith("# log channel")
+        )
```
laggron42: I'm not a fan of removing data from the config file. Just leave the variable where it is and raise a warning.

You can eventually mark the entry as deprecated[^1] in the json schema.

[^1]: https://json-schema.org/understanding-json-schema/reference/annotations

laggron42 (changes requested):
Thanks a lot for this!

I know it's tricky since Ballsdex is an async program, and logging is only synchronous, but I'm not a fan of the way the handler is made. `create_task` will return immediately, which is good to avoid blocking, but errors will never be caught, and it's still going to slow down the event loop.

The proper approach when writing handlers that take time is to use a queue, and depile queue elements in a separate thread. This is already how logging is configured globally (see [here](https://github.com/Ballsdex-Team/BallsDex-DiscordBot/blob/master/ballsdex/logging.py#L36)).\
You can read more about blocking handlers in the logging cookbook, lot of great resources and examples there: https://docs.python.org/3/howto/logging-cookbook.html#dealing-with-handlers-that-block

TL;DR:
- Remove the async logic from your handler, it will be running in a separate thread. `discord.py` already provides a [`SyncWebhook`](https://discordpy.readthedocs.io/en/latest/api.html#discord.SyncWebhook) class, removing the need to be async
- Use a QueueHandler and a QueueListener to set up the logger

---

At this point, it may be easier to declare this handler globally, for all loggers, which makes initialization very easy in `ballsdex/settings.py`, and then wait for an extra argument from the logger to know whether it should be sent to the webhook. Something like this for example:
```py
log.warning("Someone did this", extra={"send_webhook": True})
```
Another good read regarding extra data: https://docs.python.org/3/howto/logging-cookbook.html#adding-contextual-information-to-your-logging-output

---

Sorry for yapping so much, it's just that our instance of Ballsdex requires a lot of optimization so I need to be very careful. *I also really fucking love the logging library, so I'm always glad to use its features.*

PR #672, opened 2025-10-31
Pin postgres to v17

dormieriancitizen:

Pin postgres to version 17. Right now, if anyone pulls/tries to set up a new instance of ballsdex, they will be met by a scary "postgres is unhealthy" error and have to troubleshoot it—not ideal.
Unless someone has made local changes to the mountpoint, nobody should have to downgrade postgres, since v18 will not run with the current docker-compose.
Preferring v17 over v18 because I want to limit the amount of hellish postgres major version upgrades necessary.

Yes

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -70,7 +70,7 @@ services:
       "
 
   postgres-db:
-    image: postgres
+    image: postgres:17
     restart: always
     hostname: postgres
     shm_size: 1g


Dotsian:
y

HaapsaluYT:
y

AngerRandom:
y

uint128-t:
n

laggron42:
I updated the docs for the `:17` fix, which will stay until 3.0

For 2.X, there is no good solution. Forcing Postgres 17 for everyone will break many user's configuration, and changing the mount to comply with Postgres 18 will reset existing users's databases.

PR #673, opened 2025-11-06
Fix backups and transfers docs

dormieriancitizen:

Fix backup and transfers docs; previously no container was specified so it would error. Also, add --wait flag to `docker compose up` so that the DB is known to be healthy before any copying takes place (otherwise there would be errors but only sometimes, and it would work if you ran it again).

DB dump tested and worked with the new command.


dormieriancitizen:
Pretty sure this is only failing CI because 3.0-dev is in the middle of menus refactor and fails as well rn.

PR #678, opened 2025-11-23
Add server config status command

Ray-Hsueh:

Add a new command in config pack, allowing user to check whether the server has configured or not.

Newcomers are spared the temptation to tinker with channel settings, having no choice but to repeatedly restart to verify whether the configuration has succeeded.

Yes

### DIFF
--- a/ballsdex/packages/config/cog.py
+++ b/ballsdex/packages/config/cog.py
@@ -124,9 +124,10 @@ class Config(commands.GroupCog):
                         ephemeral=True,
                     )
             else:
+                config_cmd = self.channel.extras.get("mention", "`/config channel`")
                 await interaction.response.send_message(
                     f"{settings.bot_name} is now enabled in this server, however there is no "
-                    "spawning channel set. Please configure one with `/config channel`."
+                    f"spawning channel set. Please configure one with {config_cmd}."
                 )
 
     @app_commands.command()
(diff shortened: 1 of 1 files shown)


ballsdex/packages/config/cog.py
```diff
@@ -128,3 +128,33 @@ async def disable(self, interaction: discord.Interaction["BallsDexBot"]):
...
+        """
+        config = await GuildConfig.get_or_none(guild_id=interaction.guild_id)
+        if not config or not config.spawn_channel:
+            await interaction.response.send_message(
+                f"{settings.bot_name} is not configured in this server yet.\n"
+                "Please use `/config channel` to set a channel.",
+                ephemeral=False,
+            )
+        else:
+            channel = interaction.guild.get_channel(config.spawn_channel)
```
laggron42: You can make the error go away by asserting that `interaction.guild` exists.
I also added a check for the guild's availability, very rare situation but worth handling. 
```suggestion
            assert interaction.guild
            if interaction.guild.unavailable:
                await interaction.response.send_message(
                    "Your server is unavailable to the bot. Readding it may fix this."
                )
                return
            channel = interaction.guild.get_channel(config.spawn_channel)
```

laggron42 (commented):
Forgot to include it in my first comment, but can you also mention the `/config channel` command. This line will be either the string mentioning the command, or fall back to markdown if not found.
```py
config_cmd = self.channel.extras.get("mention", "`/config channel`")
```

PR #681, opened 2025-11-26
Add info about catch time

dormieriancitizen:

Adds info about catch time in the /balls info view. Also works on list, last, anything that uses prepare_for_message really.

image

This could be done in one giant ternary but I think that would look a little ugly, so even though this is more verbose I prefer it.

I don't think there's any harm in showing the users this info. Rounded to 3 decimal places, that seems sufficient to me. It shows nothing if the info isn't found or if the timedelta is broken.

Yes

### DIFF
--- a/ballsdex/core/models.py
+++ b/ballsdex/core/models.py
@@ -386,11 +386,6 @@ class BallInstance(models.Model):
             else None
         )
 
-        if isinstance(catch_time, timedelta) and not 0 < catch_time.total_seconds() < 60 * 30:
-            # some balls have a negative timedelta because of tz stuff,
-            # or if it's greater than 30 minutes, the view would have expired,
-            catch_time = None
-
         catch_time_msg = f" in {catch_time.total_seconds():.3f}s" if catch_time else ""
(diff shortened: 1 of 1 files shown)


ballsdex/core/models.py
```diff
@@ -379,9 +379,24 @@ async def prepare_for_message(
...
+            self.catch_date - self.spawned_time
+            if (self.catch_date and self.spawned_time)
+            else None
+        )
+
+        if isinstance(catch_time, timedelta) and not 0 < catch_time.total_seconds() < 60 * 30:
+            # some balls have a negative timedelta because of tz stuff,
+            # or if it's greater than 30 minutes, the view would have expired,
+            catch_time = None
+
```
laggron42: Pretty sure that can be removed entirely, I checked on Ballsdex's db and there are 0.000018% instances that don't fit these conditions, it's fine to let the input be a bit fucked (especially since `isinstance` is an expensive call)
```suggestion
```

flaree (approved):

PR #686, opened 2025-12-16
3rd party package integration

### LINKED ISSUE #676: 3rd party packages integration
## Summary

Dropping this as part of the 3.0 milestone, the bot should handle 3rd party packages as part of its build procedure.

This will *not* provide user-facing commands or scripting that automatically installs packages, instead it should just provide the core functionality that can later be extended by 3rd party developers.

## Details

With the new Docker build procedure and compose configuration, the source code isn't bind-mounted anymore to the container, it is instead installed with the virtual environment, which is read-only at runtime.\
To provide a way for 3rd-party packages to keep functioning without compromising security (which will inevitably happen if we don't provide tooling), there should be a standardized way to install discord.py extensions and Django apps, without editing the core files. Anything that requires editing files is to be considered a fork and is out of scope of this issue.

I see two solutions to this:

- Have an `extra-packages` folder which is made available in the Python path inside the container. This should still not be mounted read-write, there should be no way for an admin to edit important files during runtime.\
  This is the easiest solution, but has multiple downsides: no version control, no way to provide extra pip dependencies
- Read from a `requirements.txt` file and install them using `pip` during Docker build time.\
  I like that solution better because it allows for extra pip dependencies, proper versioning, and even the interpretation of `pyproject.toml` fields. However, this is only easy for the Docker builds, and remains to be done manually for Dockerless installs.

I will be experimenting solutions soon. If you're a 3rd party package dev and have recommendations or specific things that you'd like to see built-in, drop a comment below.

laggron42:
## Description of the changes

This defines a standardized way to integrate extra packages for a Ballsdex bot.

### User's perspective

From a user's perspective, a new file `config/extra.toml` must be created with a list of packages to add:
```toml
[[ballsdex.packages]]
location = "git+https://github.com/laggron42/awesomebattle@1.0.3"
path = "awesome_battle"
enabled = true
```

This will be read during docker build time and installed using pip. Then, it will be read at run-time to add the extra stuff to the list of django apps and dpy packages.

It is also usable without docker, but the `location` key isn't necessary and the package must be installed manually.

### Developer's perspective

A package must be placed inside its own pip-installable repository. There must be a django app, which contains a discord.py package:
```
test_app
├── pyproject.toml
└── test_app
    ├── __init__.py
    ├── admin.py
    ├── apps.py
    ├── migrations
    │   └── __init__.py
    ├── models.py
    ├── package
    │   ├── __init__.py
    │   └── cog.py
    ├── tests.py
    └── views.py
```

> [!TIP]
> Since it's a standard Python package, `pyproject.toml` can be used however you want to customize the build system: add dependencies, use special packagers, include C or Rust code using alternate builders...

Then inside `apps.py`, a new `dpy_package` attribute is added to specify where the package is located. The absence of this key results in the django app being registered without a discord.py package
```py
from django.apps import AppConfig

class TestAppConfig(AppConfig):
    name = 'test_app'
    dpy_package = "test_app.package"
 ```

Note that the `path` key of `extra.toml` allows you to have multiple apps per repository, but that will require a bit of tinkering with the package builder to allow multiple outputs.

---

One more thing is the presence of the `extra` folder in Ballsdex's repository, it's there to assist developers in coding their packages without having to publish their changes on Github all the time. Let's suppose you want to create a new package called `my_cool_app`

1. `mkdir extra/my_cool_app && cd extra/my_cool_app`
2. `git init` (and do git stuff later if you want to)
3. Create the new Django app with `django-admin startapp my_cool_app` (activate venv and `export DJANGO_SETTINGS_MODULE=admin_panel.settings`)
4. Add `dpy_package = "my_cool_app.package"` in `apps.py`
5. `mkdir my_cool_app/package` and place your usual discord.py stuff in that new folder
6. Now edit `config/extra.toml` at the root of the BD repository, and add the following
    ```toml
    [[ballsdex.packages]]
    location = "/code/extra/my_cool_app"
    path = "my_cool_app"
    enabled = true
    ```
7. `docker compose build`
8. `docker compose up --watch`
    The `--watch` option will synchronize the files inside the container, without having to rebuild or restart, you can then use the `b.reload` command as usual. If you're developing on the admin panel, it will automatically restart when needed.

Resolves #676

### DIFF
--- a/docs/dev/custom-package.md
+++ b/docs/dev/custom-package.md
@@ -433,29 +433,42 @@ Once your package is ready, you can choose to publish it!
     Visit [https://choosealicense.com/]() and put the contents of your license in a file named
     `LICENSE` at the root of your repository (`extra/my_cool_repo/LICENSE`).
 
-    !!! note
-        The contents of your LICENSE may be displayed via the `/about` command in the future.
-
 4.  Write a cool `README.md` describing your app.
 
 5.  Review the contents of your `pyproject.toml` and add some metadata. Don't forget to put the
     license.
 
-    ```toml hl_lines="6"
+    ```toml hl_lines="5-11 16 20-22"
     [project]
     name = "my_cool_repo"
     version = "1.0.0"
 
     description = "A very nice repository"
     license = "MIT"  # replace with your own license
+    license-files = ["LICENSE"]
     authors = [
         { name = "laggron42", email = "laggron42@ballsdex.com" },
     ]
     readme = "README.md"
+
+    dependencies = [
+        # you can require a specific version of ballsdex here
+        "ballsdex>=3.0.0",
+        # and extra dependencies if needed
+    ]
+
+    [project.urls]
+    # put your own links here
+    # https://packaging.python.org/en/latest/specifications/well-known-project-urls/#well-known-project-urls
+    Homepage = "https://github.com/username/repo-name"
     ```
 
     Check out the [`pyproject.toml` documentation](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#writing-pyproject-toml)
     to read more about what you can put there.
+
+    !!! info
+        The `name`, `version`, `description`, `license-files`, `authors` and `project.urls` fields
+        will be displayed to all users via the core `/about` command.
 
 6.  Commit your code.
 


dormieriancitizen (approved):
Mostly for posterity's sake—see earlier feedback on discord, but this works great for my needs.

PR #704, opened 2026-01-10
Add `/admin balls collection`

Caylies:
## Admin Balls Collection

Adds a `/admin balls collection` command which allows admins to view how many balls exist along with the ball's special count for every special.

[This command was suggested](https://discordapp.com/channels/1255250024741212262/1437116382931189933)

### Images

example1
example2

Yes (except for seeing if the trade count is accurate, but it most likely is)

### DIFF
--- a/ballsdex/packages/admin/balls.py
+++ b/ballsdex/packages/admin/balls.py
@@ -12,6 +12,7 @@ from django.db.models import Count
 from django.urls import reverse
 
 from ballsdex.core.bot import BallsDexBot
+from ballsdex.core.discord import LayoutView
 from ballsdex.core.utils import checks
 from ballsdex.core.utils.buttons import ConfirmChoiceView
 from ballsdex.core.utils.menus import ChunkedListSource, ItemFormatter, Menu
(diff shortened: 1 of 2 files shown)


flaree:
personally i think if there's to be an admin command for it then it should be in the same format as the previous one, especially given by the looks it can only fit a couple a specials on the page making it cumbersome to find what specials

Caylies:
> if there's to be an admin command for it then it should be in the same format as the previous one

The current collection embed in the balls cog doesn't support pagination and crashes if there's too many specials (the bot was tested with ~1 million ball instances and 500 specials). The admin format simply fixes that issue and provides a bit more information. ~~I did forget to add the spawn thumbnail, but I'm working on fixing that up rn~~ (EDIT: spawn thumbnail doesn't look good with this menu)

> given by the looks it can only fit a couple a specials on the page making it cumbersome to find what specials

If you want to find the count of a specific special, use the count command. I can adjust the amount of specials shown on the page to match the base collection command, as the current amount was limited so it could look better stylistically

laggron42:
Hey, first of all I really the effort you've put in this PR!

I got a bit tired of reimplementing (copying) all user commands in `/admin` just for running them on others, so instead I added `/admin impersonate` which allows to use the bot as someone else, fixing the need to reimplement all those. Thus, this PR isn't needed anymore.

I know it's not fun to have PRs closed, I'm really sorry about this...

PR #706, opened 2026-01-10
Force Terms of Service Agreement

### LINKED ISSUE #568: [bot] Accept TOS message on app command use, if no player is registered.
### Self-hosting

Official Ballsdex bot

### Description

If a user tries to run a command, and they aren't registered in the bot, instead of the regular response they would get from the app command, they get the accept TOS embed, the one currently used in /config channel.

(This is from a suggestion in the main server, https://discord.com/channels/1049118743101452329/1336299837502783619)

Caylies:
## Force Terms of Service Agreement [TESTED]

A feature suggested in the official Ballsdex server that got accepted, but not added
The guildconfig package was also refactored

By technicality, resolves #568 

### Images

tos-1

Terms of Service agreement being accepted

tos-2

Allowing command to succeed

tos-3

Guild config warning embed (condition was altered for testing)


Caylies:
I don't know why the formatter step is failing. If someone could tell me what I'm missing (because this PR was formatted with ruff using the correct configuration), then that'd be greatly appreciated :)

Ray-Hsueh:
> I don't know why the formatter step is failing. If someone could tell me what I'm missing (because this PR was formatted with ruff using the correct configuration), then that'd be greatly appreciated :)

Run ruff format --check --diff

Caylies:
> > I don't know why the formatter step is failing. If someone could tell me what I'm missing (because this PR was formatted with ruff using the correct configuration), then that'd be greatly appreciated :)
> 
> Run ruff format --check --diff

I already tried this command and it said everything was fine

Caylies:
Found the  issue, checks are passing again

ballsdex/core/bot.py
```diff
@@ -132,7 +217,19 @@ async def interaction_check(self, interaction: discord.Interaction[Bot], /) -> b
...
+
+        if not await bot.blacklist_check(interaction):
+            return False
+
+        player, _ = await Player.objects.aget_or_create(discord_id=interaction.user.id)
+
+        if not player.accepted_tos:
+            await interaction.response.send_message(
+                embed=TOS_EMBED, view=AcceptTOSView(interaction, player), ephemeral=True
+            )
```
laggron42, line 229: `interaction_check` is also called for autocompletes, you need to check the interaction type to ensure you can actually reply (see the code above).

ballsdex/core/bot.py
```diff
@@ -132,7 +217,19 @@ async def interaction_check(self, interaction: discord.Interaction[Bot], /) -> b
                 except discord.NotFound:
                     pass
             return False  # wait for all shards to be connected
-        return await bot.blacklist_check(interaction)
+
+        if not await bot.blacklist_check(interaction):
+            return False
+
+        player, _ = await Player.objects.aget_or_create(discord_id=interaction.user.id)
```
laggron42, line 224: Making a database call here is extremely expensive. `interaction_check` is called *all the time*, for every button clicked, for every letter you type in an autocompletion field.

For blacklist checks, we opted for caching, but even then I'm not convinced it's appropriate to make that check here. Some commands like `/about` shouldn't require ToS approval.

Here's my suggestion for another implementation:
- Replace all `Player.objects.aget_or_create` with `aget`
  - If the fetch happens for the player that invoked the command, let the exception propagate
  - If the fetch happens for another player (trade, give, checking someone else's list), handle the exception (as usual)
- In `on_command_error`, add an exception handler for `Player.DoesNotExist`
  - *Here* you can send the embed, or redirect to a command that lets the player accept ToS.

ballsdex/core/bot.py
```diff
@@ -107,6 +123,75 @@ async def on_request_end(
...
+
+        message = await self.interaction.original_response()
+
+        if message:
+            button.disabled = True
+
+            try:
+                await message.edit(view=self)
+            except discord.HTTPException:
+                pass
```
laggron42, line 171: `interaction.edit_original_message` does all that, you shouldn't edit a message with the bot API when you have that option (has its own ratelimits).

It needs to be called after you've responded to the interaction (or call `interaction.response.defer()`), otherwise it's going to timeout.

ballsdex/core/bot.py
```diff
@@ -107,6 +123,75 @@ async def on_request_end(
     http_counter.labels(route_key, params.response.status).observe(time)
 
 
+class AcceptTOSView(discord.ui.View):
```
laggron42, line 126: Inherit from `ballsdex.core.discord.View`, there are some additional handlers and fixed typing

ballsdex/core/bot.py
```diff
@@ -107,6 +123,75 @@ async def on_request_end(
...
+                await message.edit(view=self)
+            except discord.HTTPException:
+                pass
+
+        await interaction.response.send_message(
+            "You have agreed to the terms of service and may now use commands.", ephemeral=True
+        )
+
+    async def on_timeout(self) -> None:
+        message = await self.interaction.original_response()
```
laggron42, line 178: To avoid two HTTP calls, you can also use `interaction.edit_original_message` here

laggron42 (changes requested):

Caylies:
I won't have time to get to this for reasons

PR #710, opened 2026-01-13
Improve queryset ordering for deterministic results

Ray-Hsueh:

Fixed balls jumping around between pages. I added a secondary sort key (like -id or name) to sorting.py, the main list command, and the paginator. This stops the order from shuffling when stats or counts are the same.

[bug report](https://discord.com/channels/1049118743101452329/1460422979526852608)

No

### DIFF
--- a/ballsdex/core/utils/sorting.py
+++ b/ballsdex/core/utils/sorting.py
@@ -30,25 +30,27 @@ def sort_balls[QS: QuerySet[BallInstance]](sort: SortingChoices, queryset: QS) -
         The same queryset modified to apply the ordering. Await it to obtain the result.
     """
     if sort == SortingChoices.duplicates:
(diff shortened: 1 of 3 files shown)


Ray-Hsueh:
I believe [this](https://discord.com/channels/1049118743101452329/1455402765202231367) can also be fixed too

laggron42:
Thanks for the PR, after further testing, sorting by `-id` indeed fixes the duplicate problem.

However, I'm not going to merge this because it adds the sorting in way too many places, where the only place that needs explicit ID sorting is when pagination is invoked (list and bulk_add). I applied the fix in 650ee70b92a70eecb1dcc921a20713993b046ed9, see how shorter it is. Thanks for finding the root solution!

PR #717, opened 2026-01-17
Fix "this server" filter in /balls completion command

Ray-Hsueh:

Fixes an issue where the `this server` filter in the `/balls completion` command was being silently ignored because the `guild_id` was not passed to the filter_balls utility.

Thanks to Discord user **james_redstone** for bringing this issue up.

No


laggron42 (approved):
Good catch

PR #719, opened 2026-01-25
image_gen.py

alvinacarolinaflorencio-arch:

Yes
No


dormieriancitizen:
This looks like a PR made in error, but if not, this will definitely not work since logo is not an attribute on BallInstance

PR #726, opened 2026-02-15
Skip impersonation for admin impersonate command

### LINKED ISSUE #724: Can't unimpersonate after impersonating
### Category

Admin (/admin commands)

### Description

After impersonating a user without admin privileges on the panel, you can't unimpersonate afterwards, because the impersonated user doesn't have perms.

Image

### Steps to reproduce

`/admin impersonate <user without perms>`
`/admin impersonate`
see failure

### Expected behavior

Should unimpersonate you. It's probably fine to del a user from the impersonations table upon running the command regardless of whether they have perms to run the command for a fix.

### Screenshots

_No response_

### Version

_No response_

### Error stack trace

```python traceback

```

Ray-Hsueh:

fix #724 

Yes

### DIFF
--- a/ballsdex/core/bot.py
+++ b/ballsdex/core/bot.py
@@ -134,7 +134,10 @@ class CommandTree[Bot: BallsDexBot](app_commands.CommandTree[Bot]):
                     pass
             return False  # wait for all shards to be connected
 
-        if impersonated := impersonations.get(interaction.user.id, None):
+        if str(interaction.command) == "admin impersonate":
+            # if the user is trying to stop impersonating, don't impersonate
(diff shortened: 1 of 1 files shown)


dormieriancitizen:
Does not work for me with slash commands
image

Does work when using text commands but ideally we'd fix both.

Ray-Hsueh:
> Does not work for me with slash commands image
> 
> Does work when using text commands but ideally we'd fix both.

It should be fixed now. I've tested it and it works.

dormieriancitizen (approved):
Works correctly for me now

PR #730, opened 2026-03-09
Respect the trade cooldown option in the settings

Ray-Hsueh:

from [bug reports](https://discord.com/channels/1049118743101452329/1480029178912051340)

Yes

### DIFF
--- a/ballsdex/packages/trade/trade.py
+++ b/ballsdex/packages/trade/trade.py
@@ -530,8 +530,7 @@ class TradeInstance(LayoutView):
             remaining = 10 - elapsed
             if remaining > 0:
                 await interaction.response.send_message(
-                    f"Please wait {remaining:.0f} more second(s) before confirming.",
-                    ephemeral=True,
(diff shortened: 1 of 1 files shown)


ballsdex/packages/trade/trade.py
```diff
@@ -518,6 +519,12 @@ async def cancel_button(self, interaction: Interaction, button: Button):
     async def confirm_button(self, interaction: Interaction, button: Button):
         trader = {self.trader1.user.id: self.trader1, self.trader2.user.id: self.trader2}[interaction.user.id]
         await interaction.response.defer()
+        both_bypass = (
+            self.trader1.player.trade_cooldown_policy == TradeCooldownPolicy.BYPASS
+            and self.trader2.player.trade_cooldown_policy == TradeCooldownPolicy.BYPASS
+        )
+        if not both_bypass:
```
flaree: This isnt how it should be handled, it shouldnt auto confirm after the 10 seconds.

It should return a message saying how long left until they can confirm

flaree (changes requested):

flaree (approved):

PR #732, opened 2026-03-10
Added sort by catch_time in /balls list

cewlgruyere:
Resolves [dev server suggestion](https://discord.com/channels/1255250024741212262/1456422225681977508)
Makes it so you can sort by catch time in /balls list

Yes


admin_panel/bd_models/management/commands/startbot.py
```diff
@@ -199,7 +199,7 @@ def bot_exception_handler(bot: BallsDexBot, bot_task: asyncio.Future):
     """
     try:
         bot_task.result()
-    except (SystemExit, KeyboardInterrupt, asyncio.CancelledError):
+    except SystemExit, KeyboardInterrupt, asyncio.CancelledError:
```
flaree: Theres a bunch of random exceptions changed here, they shouldnt be touched

ballsdex/core/utils/enums.py
```diff
@@ -27,6 +27,7 @@ class SortingChoices(enum.Enum):
     attack_bonus = "-attack_bonus"
     stats_bonus = "stats"
     total_stats = "total_stats"
+    catch_speed = "catch_speed"
```
flaree:
```suggestion
    catch_time = "catch_time"
```

we normally call it catch time, so better to keep it consistent

flaree (changes requested):
Theres a bunch of random exceptions changed here, they shouldnt be touched

flaree:
Hi @cewlgruyere, just following up as you addressed one review but not the other - do you intend on fixing the issue?

cewlgruyere:
Yes, i intend on fixing it

flaree (approved):
Thanks for the PR


I added a slight change, instead of the exclude it now puts those with null spawned_time at the end, as majority of balls in ballsdex currently do not have a spawned_time so they would simply not show up and people would be confused.

PR #737, opened 2026-03-22
[admin] Readd `/admin balls create` command

Valen7440:

resolves [dev server suggestion](https://discord.com/channels/1255250024741212262/1476723177903095818)

Yes

Captura de pantalla 2026-03-21 231827
Captura de pantalla 2026-03-21 231907

### DIFF
--- a/ballsdex/packages/admin/balls.py
+++ b/ballsdex/packages/admin/balls.py
@@ -441,7 +441,7 @@ async def balls_create(
     if not emoji:
         await ctx.send(
             "The bot couldn't find the given emoji. Maybe it doesn't exist or the bot doesn't have access to it.",
-            ephemeral=True
+            ephemeral=True,
         )
         return
     await ctx.defer(ephemeral=True)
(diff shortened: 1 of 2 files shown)


dormieriancitizen (changes requested):
This will break when running without docker (`/code` isn't a valid path when not in a container), could you replace `save_file` with [django file management](https://docs.djangoproject.com/en/6.0/topics/files/)?

The only reason the method exists is because in v2 we didn't have django in the bot process

Valen7440:
> This will break when running without docker (`/code` isn't a valid path when not in a container), could you replace `save_file` with [django file management](https://docs.djangoproject.com/en/6.0/topics/files/)?
> 
> The only reason the method exists is because in v2 we didn't have django in the bot process

Ready. I replaced `save_file` with [ContentFile](https://docs.djangoproject.com/en/6.0/ref/files/file/#the-contentfile-class) that works for me.

dormieriancitizen (approved):
Works for me now

PR #742, opened 2026-04-06
Add media check command

dormieriancitizen:
Adds a owner command `{prefix}.checkmedia` for fixing broken states / identifying spawn errors.

Inspired by the hundreds of patreon dexes with broken emojis.
Also checks spawn image size / presence.

Sample:
image

- [x] Yes
- [ ] No

### DIFF
--- a/ballsdex/core/commands.py
+++ b/ballsdex/core/commands.py
@@ -12,7 +12,7 @@ from ballsdex.core.dev import send_interactive
 from ballsdex.core.discord import LayoutView, View
 from ballsdex.core.utils.formatting import pagify
 from ballsdex.core.utils.menus import Menu, TextFormatter, TextSource
-from bd_models.models import Ball, Special
+from bd_models.models import Ball
 from settings.models import load_settings, settings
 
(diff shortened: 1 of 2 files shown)


ballsdex/core/commands.py
```diff
@@ -132,6 +132,48 @@ def wrapper():
...
+        lines.append("# Missing emojis:")
+        balls = [ball async for ball in Ball.objects.all()]
+
+        for ball in balls:
+            if self.bot.get_emoji(ball.emoji_id):
+                continue
+
+            lines.append(f" - Ball `{ball.country}` has a broken emoji")
+
+        specials = Special.objects.all()
```
flaree: Special emojis can only be unicode, so i dont think they need to be checked

flaree (dismissed):

ballsdex/core/commands.py
dormieriancitizen: The special helptext says that I can use an emoji ID, but I just tried it and it breaks things, so I'll fix that as well

flaree (approved):

PR #743, opened 2026-04-06
Validate group names

dormieriancitizen:

Collectible name and plural name must match the discord command schema, because they are used in commands. If they don't, tree sync fails. So validate the settings entries with the same validator as slash command name. 

- [x] Yes
- [ ] No


admin_panel/settings/migrations/0006_alter_settings_admin_channel_ids_and_more.py
```diff
@@ -0,0 +1,53 @@
...
+                blank=True,
+                default="",
+                help_text="Semicolon-delimited channel IDs where staff may bypass inventory privacy. Ignored for owners.If empty, then admin commands can be used everywhere.",  # noqa: E501
+                validators=[
+                    django.core.validators.RegexValidator(
+                        re.compile("^(\\d{17,21}(;\\d{17,21})*)?$"), message="The IDs must be semicolon-separated"
+                    )
+                ],
+            ),
+        ),
```
laggron42, line 26: Why is that here, was a previous migration missed?

dormieriancitizen: I guess so; it must have appeared when I made migrations

flaree (approved):

PR #752, opened 2026-05-03
[admin] update permission checks for blacklist and info commands

flaree:

update permission checks for blacklist and info commands


dormieriancitizen (approved):
Correctly blocks commands for unpermed users for me (and works for users with perms)

PR #756, opened 2026-05-07
[about] Fix "3rd party packages" not working when u click a package with hyphens in their name

hiboman:
this fixes the error that would occur if a package had hyphens in its name
- [x] Yes
- [ ] No

### DIFF
--- a/ballsdex/packages/info/license.py
+++ b/ballsdex/packages/info/license.py
@@ -16,7 +16,8 @@ packages = importlib.metadata.packages_distributions()
 extra_apps_dist: dict[str, importlib.metadata.Distribution] = {}
 for app in settings.EXTRA_APPS:
     try:
-        extra_apps_dist[app] = importlib.metadata.distribution(packages[app][0])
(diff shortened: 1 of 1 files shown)


ballsdex/packages/info/license.py
```diff
@@ -16,7 +16,7 @@
 extra_apps_dist: dict[str, importlib.metadata.Distribution] = {}
 for app in settings.EXTRA_APPS:
     try:
-        extra_apps_dist[app] = importlib.metadata.distribution(packages[app][0])
+        extra_apps_dist.update({(dist := importlib.metadata.distribution(packages[app][0])).name: dist})
```
laggron42: The walrus syntax is making this poorly readable, using two lines here is fine.
```suggestion
        dist = importlib.metadata.distribution(packages[app][0])
        extra_apps_dist[dist.name] = dist
```

laggron42 (changes requested):
Thanks for the contribution, just a little style change needed

laggron42 (approved):

PR #757, opened 2026-05-07
[admin panel] Fix "add trade" page

hiboman:
this fixes the error that would occur if you try to click add trade in the admin panel
- [x] Yes
- [ ] No

### DIFF
--- a/admin_panel/templates/admin/bd_models/trade/change_form.html
+++ b/admin_panel/templates/admin/bd_models/trade/change_form.html
@@ -1,6 +1,7 @@
 {% extends "admin/change_form.html" %}
 
 {% block after_related_objects %}
+{% if original %}
 <div class="tabular inline-related" style="bottom:15px">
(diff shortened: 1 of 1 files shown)


laggron42 (commented):
Thanks for the contribution! There are a few missing things in your patch, but if you want to make the "Add trade" page useable, the whole table displaying the trade list is meaningless, it would be better to wrap the entire table in a condition. Django provides `original` which is only defined on the change view, so you could do this instead:
```py
{% block after_related_objects %}
{% if original %}
<div class="tabular inline-related" ...>
  ...
</div>
{% endif %}
```

hiboman:
Thank you, will commit those changes

laggron42 (approved):
I am not exactly sure why you want this (trade items aren't available in the "add" view, and adding trade entries isn't going to change ownerships, kinda risky), but there's no real reason to not add this fix

PR #758, opened 2026-05-07
Fix libraqm versioning

dormieriancitizen:

Change `edge` to `v3.23` because libraqm-dev has moved off edge. We can't just bump alpine yet because python isn't out for alpine 3.23.

- [x] Yes
- [ ] No


dormieriancitizen:
Actually, now I'm confused, because while this does work, edge still has libraqm so it shouldn't be needed

dormieriancitizen:
Yes, I just checked and it builds cacheless without—idk how bloxy triggered this error

PR #763, opened 2026-05-18
Doc improvements: add multiple bot host guide & use management cmd for yml import

dormieriancitizen:

Add a guide for multiple bot hosting, and switch to using the management command (rather than importing a file from the panel) for v3 upgrades since it's a little cleaner and doesn't require a local admin account if you're already using oauth2.

They're docs. But the commands in question are the ones I use to manage patreon hosts, so they should work.

### DIFF
--- a/docs/selfhosting/misc/hosting-multiple-bots.md
+++ b/docs/selfhosting/misc/hosting-multiple-bots.md
@@ -1,0 +1,36 @@
+Hosting multiple bots on one machine is very similar to hosting one.
+
+First, enter a Linux terminal (WSL or proper linux).
+
+## 1. Download the bot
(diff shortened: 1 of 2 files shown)


docs/selfhosting/misc/hosting-multiple-bots.md
```diff
@@ -0,0 +1,36 @@
+Hosting multiple bots on one machine is very similar to hosting one.
+
+First, enter a Linux terminal (WSL or proper linux).
+
+## 1. Download the bot
+Clone Ballsdex. **You must replace <botname>** with some text you haven't already used for another bot (the dex's name is a decent choice)
```
flaree, line 6: Maybe a disclaimer that renaming it will 'lose' data, keep it something that shouldnt change?

flaree (approved):

PR #772, opened 2026-06-03
[player friend list] allow impersonators to remove friends

hiboman:
Allows an impersonating admin to unfriend someone via the impersonated users friend list

- [x] Yes
- [ ] No

### DIFF
--- a/ballsdex/packages/players/views.py
+++ b/ballsdex/packages/players/views.py
@@ -8,6 +8,7 @@ import discord
 from discord import ButtonStyle, SeparatorSpacing
 from discord.ui import ActionRow, Button, Container, Label, Section, Select, Separator, TextDisplay, Thumbnail, button
 
+from ballsdex.core.bot import impersonations
 from ballsdex.core.discord import Modal
 from ballsdex.core.utils.buttons import ConfirmChoiceView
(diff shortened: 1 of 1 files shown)


dormieriancitizen:
The whole point of impersonation is that it shouldn't require any downstream changes

I'm not sure why what we have now doesn't work; interaction_check should run before button presses, but solving it this way defeats the idea behind impersonation

hiboman:
> The whole point of impersonation is that it shouldn't require any downstream changes
> 
> I'm not sure why what we have now doesn't work; interaction_check should run before button presses, but solving it this way defeats the idea behind impersonation

this should do the trick though

laggron42:
Thanks for noticing this bug, indeed the impersonator isn't set properly on newly created views.

However, you only fixed this locally, for a single part of the code, *everything* in the bot using a view or a modal would need to be fixed like you did, and this is not how impersonation is meant to be used. Impersonation should happen on a global scale, and resulting code should never have to consider them. I will be fixing that on a global scale, but thank you for reporting the bug!

hiboman:
Oh, thank you for seeing this
Im always happy to report bugs ☺️

PR #773, opened 2026-06-03
[image gen] allow overlays onto cards while drawing

tobyers:

I've updated the `image_gen.py` functionality so it allows for overlays onto the collector_card section of the image, this allows for cooler, 3D designs which can include different light effects, filters, or other overlayed features.

image

image

- [X] Yes
- [ ] No

## To test:

- [ ] See if all variables are set correctly
- Not all the variables are required for the base bot
- [ ] Test to see if images are drawn at correct size
- [ ] Check that regular non-overlayed images still function normally

### DIFF
--- a/ballsdex/core/image_generator/image_gen.py
+++ b/ballsdex/core/image_generator/image_gen.py
@@ -12,11 +12,6 @@ if TYPE_CHECKING:
 
 
 SOURCES_PATH = Path(os.path.dirname(os.path.abspath(__file__)), "./src")
-WIDTH = 1500
-HEIGHT = 2000
-
-RECTANGLE_WIDTH = WIDTH - 40
-RECTANGLE_HEIGHT = (HEIGHT // 5) * 2
 
 CORNERS = ((34, 261), (1393, 992))
 artwork_size = [b - a for a, b in zip(*CORNERS)]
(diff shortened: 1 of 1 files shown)


flaree:
This drastically changes the defaults of the generator, it should be done in a way that if there's no changes then someone wouldn't notice. also it saves back to PNG which you shouldn't due - keep it as WEBPAGE due to its compression

dormieriancitizen:
Perhaps I'm biased, but I would prefer #740 and then these sorts of changes (which are cool) could be easily be implemented by a renderer in a third party package

tobyers:
> This drastically changes the defaults of the generator, it should be done in a way that if there's no changes then someone wouldn't notice. also it saves back to PNG which you shouldn't due - keep it as WEBPAGE due to its compression

I didn’t realise how many things actually changed from the base bot - I built this feature for the MonsterDex bot which already had some changes from the original BallsDex image generator code, it’s really only 1 block that would need changing. I’ll look over and compare this to the original image_gen.py code and attempt to make a more coherent version 👍

tobyers:
> Perhaps I'm biased, but I would prefer #740 and then these sorts of changes (which are cool) could be easily be implemented by a renderer in a third party package

It’s really not a massive change, I think I made it seem like more than it is accidentally lmao

I plan on revising the code and re-releasing a better version to be more easily compatible

flaree:
Closing as we have the variants PR

laggron42:
Still a huge thanks for your contribution, I know it's frustrating to have a PR closed, but it's just because we had duplicates, feel free to contribute again

PR #774, opened 2026-06-04
Fixed the crash that would happen when using /trade history days:number

cewlgruyere:
Currently the /trade history command crashes when using the days parameter, this PR fixes that

- [x] Yes
- [ ] No


laggron42 (approved):
Thanks! I have no idea how this wasn't spotted earlier

PR #778, opened 2026-06-08
fixed the trade history cog

Gloxiedev:

fixed the history cog bug

and this want tested locally

### DIFF
--- a/ballsdex/packages/trade/cog.py
+++ b/ballsdex/packages/trade/cog.py
@@ -268,7 +268,8 @@ class Trade(commands.GroupCog):
             if trade_user:
                 p2 = await Player.objects.only("id").aget(discord_id=trade_user.id)
         except Player.DoesNotExist:
-            await interaction.response.send_message("One of the players does not exist.", ephemeral=True)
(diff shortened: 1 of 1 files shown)


ballsdex/packages/trade/cog.py
```diff
@@ -268,7 +268,8 @@ async def history(
             if trade_user:
                 p2 = await Player.objects.only("id").aget(discord_id=trade_user.id)
         except Player.DoesNotExist:
-            await interaction.response.send_message("One of the players does not exist.", ephemeral=True)
+            send = interaction.followup.send if interaction.response.is_done() else interaction.response.send_message
```
flaree, line 271: You don't need this conditional, it will always be a followup.send as it's deferred straight at the beginning of the method

flaree (changes requested):

ballsdex/packages/trade/cog.py
Gloxiedev: hmm, okay

Gloxiedev:
- [ ] 

- [ ]

PR #780, opened 2026-06-08
Add balls details command

hiboman:

This adds in a balls details command that shows you information about a specific enabled countryball
And no it is nothing like balls info

- [x] Yes
- [ ] No


flaree:
Thanks for the PR, however I don't see the need for this. All the info is obtainable with balls info, a dedicated command isn't required. It would just be there for the sake of it

Closing

hiboman:
This command lets you view info of balls you dont own, so it doesn't exactly work like balls info.

flaree:
Then it's a command that I definitely wouldn't have added

PR #790, opened 2026-07-01
Add group model

flaree:

Adds group mode, lets you add balls to groups and filter based on groups

- [x] Yes
- [ ] No


admin_panel/bd_models/models.py
```diff
@@ -43,6 +43,7 @@ def image_display(image_link: str) -> SafeText:
 regimes: dict[int, Regime] = {}
 economies: dict[int, Economy] = {}
 specials: dict[int, Special] = {}
+group_balls: dict[int, set[int]] = {}
```
dormieriancitizen: With this model, I'm pretty sure it's impossible to get the group name (and other attributes if we want to add them later) without a DB call if you want to use caching, which seems important for a bunch of commands

It would make more sense to me to have `group_balls` be the same as the rest of the cached attributes, eg `dict[int, BallGroup]` and then have `BallGroup` have a cached_property `balls: list[Ball]`

dormieriancitizen (commented):
The code works for me as written, but I have a concern about the data layout for caching

admin_panel/bd_models/models.py
flaree: Pushed those changes, note I modified the migration to rename the field to avoid some dual object naming. No need for 2 migrations in the one PR so squashed them.

dormieriancitizen (approved):

PR #793, opened 2026-07-03
Add balls drop

flaree:

Add balls drop command

### DIFF
--- a/ballsdex/core/utils/transformers.py
+++ b/ballsdex/core/utils/transformers.py
@@ -164,7 +164,7 @@ class BallInstanceTransformer(ModelTransformer[BallInstance]):
         return super().get_queryset().prefetch_related("player")
 
     async def get_from_pk(self, value: int) -> BallInstance:
-        return await self.get_queryset().prefetch_related("player", "trade_player").aget(pk=value)
+        return await self.get_queryset().prefetch_related("player", "trade_player", "ball").aget(pk=value)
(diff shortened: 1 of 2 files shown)


Renzygame:
Huge issue here, if you have a package specifically the labelmaker, it will cause an error for the drop ball command to fail.

flaree:
> Huge issue here, if you have a package specifically the labelmaker, it will cause an error for the drop ball command to fail.

Then fix labelmaker, that is not our concern.

Renzygame:
It's not my package tho, unless there's also a way to pr packages then maybe I can.

dormieriancitizen:
Not only is this not the right place (use the [package repo](https://github.com/dormieriancitizen/bd-labelmaker/)), it's also not even an accurate report.

image

Renzygame:
> Not only is this not the right place (use the [package repo](https://github.com/dormieriancitizen/bd-labelmaker/)), it's also not even an accurate report.
> 
> image

Ok let me explain more in depth then if it's not accurate to the report, basically the Drop command is broken because of a package, now was it because of the package? No because if you try it in a normal spawn no drop command it would work but with the drop command it won't work but my theory might be because of the view of when the ball spawns using the drop command.

dormieriancitizen:
Not the place.

Renzygame:
> Not the place.
Okay.

PR #829, opened 2026-07-23
Add /admin promptmessage create

HaapsaluYT:

Creates a new command that allows users to make new spawn messages from Discord, rather than accessing the admin panel.

Yes, although the permissions are most likely wrong, as I'm not familiar with what perms are allocated to prompt messages.


HaapsaluYT:
I'll also note that I'm working on /admin promptmessage delete, but I haven't quite gotten it to work yet - see the half baked changes I rolled back which were leftovers from when I tried to upload the stuff on my computer

flaree:
Personally i'm not a fan of using a sub command slot for this. To me this is something that should only be done via the admin panel

PR #854, opened 2026-09-22
Reduce memory usage in convert_media.py

Nipatsuplayer:
Convert files one by one instead of passing the whole media collection to a single ffmpeg process.

Currently command generates single ffmpeg process with all media files as inputs. For a bigger image collection this will consume a lot of memory and oom might kill the process on smaller servers. Converting the files one by one uses less memory. Its little bit slower but should make the command more stable.

- [x] Yes
- [ ] No


admin_panel/media_management/management/commands/convert_media.py
```diff
@@ -99,21 +90,41 @@ def convert_media(self, *args, **options):
...
-                finally:
-                    self.stdout.write(f"ffmpeg did not complete successfully: error: {result.stderr}")
+                    raise CommandError(
+                        f"ffmpeg exited with non-0 exit code "
+                        f"{result.returncode} while converting "
+                        f"{input_file.name}!\n\n"
+                        f"{result.stderr}"
+                    )
 
+                shutil.copy2(temp_output, media_path / output_file.name)
```
dormieriancitizen: This copy action is supposed to be atomic; no files should be moved if any conversion fails. So, you should keep the tmp_dir copytree even if the conversion happens 1 by 1.

dormieriancitizen (changes requested):
I think this is good (though ffmpeg has a non-negligible start time, so maybe it would be good to include a batch flag to determine how many files are processed by each ffmpeg command?)

In terms of the code you had, I just submitted one change.

